PR #17, opened 2024-04-17
Push for changes

rphumulock:
- Fixed some styles - might want to check them.
- modified Getting getting started page.
- moved backend examples to examples route.
- moved things to von deepa (can rename).


delaneyj:
:shipit:

PR #45, opened 2024-05-18
fix problem with sets in IsLoading backend plugin

Superpat:
We had basically a race condition where the signal store was being reset in between the start and end of the fetch so the `s` was no longer pointing to the right store. We also had an issue with sets having weird interactions with signals and always setting the empty object instead of the empty set, so we switched to arrays.


delaneyj (approved):

gedw99:
looks like something in this broke the GUI ?

https://github.com/delaneyj/datastar/issues/46#issue-2305265751

Issue #56, opened 2024-05-22
need to support expressions with newlines, not just semicolons


delaneyj:
```ts
data-on-raf="
	const currentTimeMs = Date.now();
	const delta = currentTimeMs - window.lastRenderTime;
	window.lastRenderTime = currentTimeMs;
	if($timeControls.isPlaying) {
		const deltaWithMultiplier = delta * $timeControls.timeMultiplier;
		const newTime = $timeControls.currentTimeMs + deltaWithMultiplier;
		const clampedTime = Math.max($timeControls.minTimeMs, Math.min($timeControls.maxTimeMs, newTime));
		$timeControls.currentTimeMs = clampedTime;
	}
"
```
works
```ts
data-on-raf="
	const currentTimeMs = Date.now()
	const delta = currentTimeMs - window.lastRenderTime
	window.lastRenderTime = currentTimeMs
	if($timeControls.isPlaying) {
		const deltaWithMultiplier = delta * $timeControls.timeMultiplier
		const newTime = $timeControls.currentTimeMs + deltaWithMultiplier
		const clampedTime = Math.max($timeControls.minTimeMs, Math.min($timeControls.maxTimeMs, newTime))
		$timeControls.currentTimeMs = clampedTime
	}
"
```

delaneyj:
incorrectly closed by me

PR #61, opened 2024-05-23
Signal event type

### LINKED ISSUE #55: add a store SSE event to only effect the signals
Something like
```
event: datastar-store
data: merge patch
data: patch {foo:'bar'}
```
where you can update the store directly without DOM fragments

delaneyj:
fixes #55 
makes local signals with _prefix


delaneyj:
Also fixes #56

PR #120, opened 2024-08-19
#119 Fix for the style being repeatedly appended from repeated events bein…

### LINKED ISSUE #119: Show Plugin duration transition styles applied every SSE event on child elements
When adding .duration modifier to an element using the show plugin, if that element has a child that fires multiple SSE messages, the style for the show transition gets applied multiple times.

A chatter in discord "Flister" pointed this out.

> Flister
> Hi,
> Testing using the backend examples node.js and python, and adding duration to data-show
> <div data-show.duration="$show">
> For each update from the feed this style-block gets added to the <head>

rphumulock:
Fixes #119


delaneyj:
Is there a test for this to make sure we don't have regressions going forward?

PR #139, opened 2024-09-24
fix: display actions only when editing a row

blinkinglight:


delaneyj:
Sorry I fixed before I saw this PR.  Thanks though!

PR #140, opened 2024-09-25
nix shell init

phanirithvij:
Added a nix shell, useful for nix users. Doesn't affect any normal functionality.

- npins init
- tools required (node, pnpm etc.)
- shellhook for playwright
- skip downloading browsers if using nix (nix downloads playwright browsers)

I've tested this in Nixos like `nix-shell` then `task` and `task test`

todo
- [x] test if it works outside nixos with just nix


phanirithvij:
abandoned effort. apologies any nix users use this simple shell.nix

```nix
let
  pkgs = import <nixpkgs> { };
in
pkgs.mkShellNoCC {
  env = {
    PLAYWRIGHT_BROWSERS_PATH = "${pkgs.playwright-driver.browsers}";
    PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS = true;
  };
  shellHook = ''
    git lfs install --local
    git lfs fetch
    git lfs checkout

    task -a
  '';
  packages = with pkgs; [
    git-lfs
    go
    go-task
    templ

    killall
    pnpm
    nodejs-slim_20
  ];
}
```

reminder to **NOT** run `pnpm playwright install` since the browsers are managed by nix.

PR #147, opened 2024-10-08
Fixed minor website examples bug ">" --> ">=" and changed minimum character number to 2

alecgargett:
Fixed minor bug ">" --> ">=" in the inline-validation example on the website and changed the minimum character number to 2.


delaneyj (approved):
:shipit:

PR #157, opened 2024-10-17
Remove merge options covered by idiomorph

potatotech:
Use cases for `inner`, `outer`, and `upsert_attributes` can all be handled by `morph` and signals.


bencroker:
Can you please explain?

delaneyj:
@potatotech  upsert_attributes isn't covered by idiomorph

potatotech:
Restored `upsert_attributes`

delaneyj:
Needs to be removed from `datastar.go` and any examples as well but getting close!

bencroker:
I think I’ve changed all instances in all examples now.

bencroker:
`main` has been merged, fyi.

delaneyj:
LGTM will put off until v20 release

bencroker:
> Use cases for `inner`, `outer`, and `upsert_attributes` can all be handled by `morph` and signals.

@potatotech can you please help me understand how `inner` can be handled when the default morph style used is `outerHTML`, which is not currently configurable in Datastar?

bencroker:
If it turns out that the only redundant merge strategy is `outer`, then I’d opt for leaving it in, as it still offers the benefit of allowing you to target an element by selector without requiring matching IDs.

bencroker:
I’ve come to the strong conclusion that there is no upside whatsoever in removing merge options. Even `outer` can be useful considering that the back-end should have full control over deciding what DOM elements to swap. Without it, an ID must already exist in the front-end. I vote _not_ to merge.

potatotech:
The backend *is* the frontend with hypermedia. If there's a missing id, add it in the template. When swapping fragments, datastar needs to target a specific element, and the only unique selector is id. Morph is sufficient because all other selectors are less specific\*.

\* There are things like `#id[attribute]` or `#id.class`, but id is already unique, so these are not necessary. You definitely should not be using them to store state either.

delaneyj:
I'm going to leave it @potatotech, while I :100: agree you should always use the idiomorph I can see value in the effiency of inner/outerhtml for sheer speed in certain situatins.  Thank you for your effort!

PR #158, opened 2024-10-17
Fix console event

bencroker:
Fixes the console event by splitting on a space and then joining the rest.


bencroker:
Fixed the fix ☺️

delaneyj (approved):
:shipit:

PR #160, opened 2024-10-17
Add `error` console mode

bencroker:
This PR adds the `error` console mode, since most browsers allow filtering by it.

I’m also curious as to the reason for having `group` and `groupEnd` there? I can’t say I’ve ever used them.

**Chrome:**

screenshot-rV76WpQc@2x

**Firefox:**

screenshot-q6ATxVdh@2x


delaneyj (approved):
:shipit:

PR #161, opened 2024-10-18
Rename `delete` event to `remove`

bencroker:
This PR renames the `delete` SSE event to `remove`.

Reasoning:
- It better describes the action.
- The event ultimately calls `target.remove()` so easier to follow in the source code.
- More clearly distinguishes the event from the `$$delete()` action plugin.


delaneyj:
needs all the examples / datastar.go updated as well but I'm keen on the idea for v20

bencroker:
I think I got them all!

bencroker:
Merged the `main` branch into this one, in case you’re still (hopefully) open to renaming the event.

delaneyj:
This is a good idea (being semantically correct).  But think it actually doesn't match now....
`datastar-delete` now supports
1. Fragment selectors
2. Store paths 
One uses `remove` the other `delete`.  Since it's already delete I'm gonna err on leaving as is instead of introducing a breaking change.  We can talk move about it on discord

bencroker:
Whether you remove a fragment from the DOM or remove a value from the store, “remove” is still more semantically correct than “delete”, in my mind. 

I’ve updated the necessary files and am reopening for further consideration.

bencroker:
My last ditch effort at convincing you is that using “remove” here helps disambiguate from the “delete” HTML verb used in the back-end plugin when source diving or searching.

delaneyj:
fixing in starfederation/datastar repo for now

PR #167, opened 2024-10-19
Change `vt` to `view-transition` and add a duration

bencroker:
This PR changes the fragment event option `vt` to `view-transition` and turns it into it a duration. If set to `0`, view transitions are not applied, otherwise they are using the provided duration.

### DIFF
--- a/packages/library/src/lib/plugins/backend.ts
+++ b/packages/library/src/lib/plugins/backend.ts
@@ -176,7 +176,7 @@ async function fetcher(method: string, urlExpression: string, ctx: AttributeCont
             currentDatatype = ''
 
           for (let i = 0; i < lines.length; i++) {
-            let line = lines[i]knownEventTypes
+            let line = lines[i]
             if (!line?.length) continue
 
             const firstWord = line.split(' ', 1)[0]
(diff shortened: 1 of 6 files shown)


packages/library/src/lib/plugins/backend.ts
```diff
@@ -439,7 +439,7 @@ export function mergeHTMLFragment(
 
         const revisedHTML = modifiedTarget.outerHTML
 
-        if (originalHTML !== revisedHTML) {
+        if (originalHTML !== revisedHTML && settleTime > 0) {
```
bencroker: @delaneyj I added a condition to only add/remove the settling class if the settle time is non-zero. Please review if you’re ok with this.

delaneyj:
need all examples and helpers updated too!

bencroker:
This is ready for review. Please carefully check my work in [backend.ts](https://github.com/delaneyj/datastar/pull/167/files#diff-7afc62c866ea4a0c5ba0c1f11c8d901fae4f1193ff4339fa0bece49f85e822db).

bencroker:
Since all that’s required for View Transition animations is CSS, I no longer think animation duration is something Datastar should handle. We’d only be allowing setting the duration, not any of the other animation effects, so my feeling is that it’s probably better to just allow enabling/disabling View Transition and leave the rest to the user, as is the case now. 

I’d still like to see `vt` renamed to `view-transition`, but understand that it is a breaking change in the name of ergonomics.

PR #171, opened 2024-10-20
[WIP] JS: Split autoloading code for script tag usage from library code

pdf:
This is probably not the best/right way of doing this, but it looks like vite does not have usable support for multiple outputs. I don't know enough about the current JS bundler landscape to know if there's a better option - I've just been using bun and tailwind directly for my personaly projects. The code changes are relatively sane I think, but the bundler side I'm not so sure about.

In this proposed scenario, all script tags would be updated to use `datastar.autoload.js`, and the library would remain at `datastar.js`. Since CDN scripts are versioned, this doesn't seem too terrible, since the script src for users would only need to change when moving to a new version, but I think this could be flipped the other way, so that the autoload version stays at `datastar.js` if `package.json` is updated.

### DIFF
--- a/packages/library/src/lib/index.ts
+++ b/packages/library/src/lib/index.ts
@@ -17,6 +17,12 @@ import { HelperActions } from './plugins/helpers'
 import { elemToSelector } from './plugins/core'
 import { Actions, AttributePlugin, DatastarEvent, datastarEventName, SendDatastarEvent } from './types'
 
+declare global {
+  interface Window {
+    ds: Datastar
+  }
+}
+
 export { apply } from './external/ts-merge-patch'
 
 export function runDatastarWith(actions: Actions = {}, ...plugins: AttributePlugin[]) {
@@ -44,8 +50,6 @@ const datastarDefaultEventOptions: CustomEventInit = {
   composed: true,
 }
 
-const winAny = window as any
-
 export const sendDatastarEvent = ((
   category: 'core' | 'plugin',
   subcategory: string,
@@ -54,7 +58,7 @@ export const sendDatastarEvent = ((
   message: string,
   opts: CustomEventInit = datastarDefaultEventOptions,
 ) => {
-  winAny.dispatchEvent(
+  globalThis.dispatchEvent(
     new CustomEvent<DatastarEvent>(
       datastarEventName,
       Object.assign(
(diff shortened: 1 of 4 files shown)


packages/library/src/lib/index.ts
```diff
@@ -74,13 +78,13 @@ export const sendDatastarEvent = ((
   )
 }) satisfies SendDatastarEvent
 
-// Timeeout allows inspector to attach to all elements before sending the event
-if (!winAny.ds) {
+export function runForInspector(fn: () => Datastar) {
+  // Timeeout allows inspector to attach to all elements before sending the event
```
anastygnome, line 82: looks like there's a typo in timeout

pdf: That typo is pre-existing, I just moved it. Can fix it though.

delaneyj:
I haven't been ignoring... In fact it spurred me to try a whole new approach.  https://github.com/starfederation/datastar  is my attempt to make it much easier to make your own "bundles" of Datastar.  This approach uses esbuild as a Go library directly.  No vite, no node_modules, just TS to JS with source maps.  Would love to here y'all's thoughts.

pdf:
Sorry for the delay in responding, I've been thinking about this.

I think for library usage, we should probably be publishing the full module path structure so that users can import specific modules, e.g. (based loosely on the `starfoundation/datastar` repo structure):

```javascript
import { Datastar } from "@starfederation/datastar";
import { PersistAttributePlugin } from "@starfederation/datastar/plugins/storage";

Datastar.load(PersistAttributePlugin);
```

Exposing the full module structure will also make it easier to implement plugin hooks or other extension patterns, for example to support https://github.com/delaneyj/datastar/discussions/166#discussion-7340484 we'd probably want a singleton in the backend module, which might be referenced something like (assume `BackendRegistry` is a hypothetical singleton that allows registering a new SSE action handler):

```javascript
import { Datastar } from "@starfederation/datastar";
import { PersistAttributePlugin } from "@starfederation/datastar/plugins/storage";
import { BackendRegistry, GetSSEActionPlugin } from "@starfederation/datastar/plugins/actions/backend";

BackendRegistry.register("custom-action", (data) => {});
Datastar.load(PersistAttributePlugin, GetSSEActionPlugin);
```

Exporting the module structure allows exposing this sort of functionality without having to hoist everything public up to the top-level namespace.

I've spent some time looking into what's required to get this working, and it looks like we'd need build one bundled output for the browser, and one set of transpiled output for the full TS source tree, with exports defined for each module boundary in `package.json` (we'll probably want to standardize on exporting aggregates from an index.ts for each directory to simplify the export definitions so they can be declared with wildcards). We should also generate type definitions so library consumers can use TS with type-safety, but it looks like this will necessitate the use of `tsc` as `esbuild` doesn't appear to be capable of generating type defs.

The above approach will work for consuming the lib from node at least - if we wanted to produce valid ESM output directly we'd need to change all the TS imports to use explicit `.js` filenames due to Node/TypeScript/ESM being bonkers. Certainly doable, but makes the TS a bit uglier. If we don't do this then consumers of the library must run their own bundler to produce valid ESM/JS for the browser, but that's likely the common approach anyway. Thoughts?

If this makes sense at all to you @delaneyj I'll take a stab at working this up against the `starfoundation/datastar` repo.

delaneyj:
Please take a look at the new repo and hop on Discord to discuss

PR #172, opened 2024-10-20
Fix `data-ref` processing

bencroker:
This PR fixes a bug with processing `ref`s.

```html
<div data-ref="hint">This is a helpful hint.</div>
<button data-on-click="~hint.remove()">Reset</button>
```

Clicking the button results in this error:

```
Error evaluating Datastar expression:
  return document.querySelector(ctx.store()._dsPlugins.refs.hint.remove)()

Error: document.querySelector(...) is not a function
```

Whereas this works:

```html
<div data-ref="hint">This is a helpful hint.</div>
<button data-on-click="(~hint).remove()">Reset</button>
```

I’m not sure what the best way to fix this is, but @delaneyj suggested wrapping in parentheses in the code to avoid forcing users to have to do it, which is what this PR does.

### DIFF
--- a/packages/library/src/lib/plugins/core.ts
+++ b/packages/library/src/lib/plugins/core.ts
@@ -42,7 +42,7 @@ const ActionProcessor: Preprocessor = {
   },
 }
 
-// Replacing ~foo with ctx.refs.foo
+// Replacing ~foo with ctx.refs.(foo)
 const RefProcessor: Preprocessor = {
   regexp: wholePrefixSuffix('~', 'ref', ''),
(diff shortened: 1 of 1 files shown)


bencroker:
I marked this PR as ready to review too soon, still WIP.

delaneyj:
Is this ready?

bencroker:
No, still WIP.

bencroker:
Assuming `ref`s _cannot_ be nested, this fixes the issue. Ready for your review, @delaneyj.

delaneyj (approved):

Issue #196, opened 2024-10-29
fix docs or rename onlyIfMissing


blinkinglight:
the code: https://github.com/delaneyj/datastar/blob/main/packages/library/src/lib/plugins/backend.ts#L264
the docs: https://github.com/delaneyj/datastar/blob/main/backends/go/site/static/md/reference/plugins_backend.md ( ifmissing false )

bencroker:
Thanks for reporting this! Fixed in https://github.com/delaneyj/datastar/pull/198.

PR #197, opened 2024-10-29
Fix `onlyIfMissing` docs

blinkinglight:
https://github.com/delaneyj/datastar/blob/main/packages/library/src/lib/plugins/backend.ts#L264


blinkinglight:
#196

bencroker:
Thanks for this. I’ve created https://github.com/delaneyj/datastar/pull/198 which fixes the code, rather than the docs. @delaneyj can decide which PR to merge.

delaneyj (approved):

PR #230, opened 2024-11-23
Fix With/Without ViewTransitions merge fragment option switcharoo

Napam:
Looks like there has been a switcharoo in between WithViewTransitions and WithoutViewTransitions merge fragment option thing

### DIFF
--- a/code/go/sdk/fragments-sugar.go
+++ b/code/go/sdk/fragments-sugar.go
@@ -58,13 +58,13 @@ func WithSelectorID(id string) MergeFragmentOption {
 
 func WithViewTransitions() MergeFragmentOption {
 	return func(o *MergeFragmentOptions) {
-		o.UseViewTransitions = false
+		o.UseViewTransitions = true
 	}
 }
 
 func WithoutViewTransitions() MergeFragmentOption {
 	return func(o *MergeFragmentOptions) {
-		o.UseViewTransitions = true
+		o.UseViewTransitions = false
 	}
 }
 


delaneyj:
Yep, we change the defaults.  It was very clear it caused too many issues with newcomers.  When a VT starts it locks out user interaction.  If you start streaming down events you can lock yourself out.  I liked the default but clear it's confusing to most.

Thanks for the PR and sorry if the flip has caused confusion

Napam:
@delaneyj 

Wait, so If I wanna explicitly specify that I **want to have** ViewTransitions, I gotta write that I want to run "**Without**ViewTransitions"?

```golang
// This explicitly activates ViewTransitions (without my changes)
NewSSE(w, r).MarshalAndMergeSignals(store, WithoutViewTransitions())
                      // It says "without" ^^^^^^^, 
```

Are you sure this is right? It sounds quite opposite to me.  Wouldn't one expect?:

```golang
NewSSE(w, r).MarshalAndMergeSignals(store, WithViewTransitions())
```

Also vice-versa for when you want to explicitly specify that you don't want ViewTransitions.

UPDATE:

I think the change in the defaults is fine, but isn't this part of the code irrelevant to the defaults? I want to explicitly override here. The defaults are in `fragments.go` line 10 and 51 no?

** Also added another comment to pseudo code above

delaneyj:
It should be WithViewTransitions(), if that doesn't work it's a bug

Napam:
So I just tested it again

In my case it looks like a bug right now. Here is my code:
```golang
func handleSSRselector(w http.ResponseWriter, r *http.Request, store *Store) {
	store.SSRSelectorShow = !store.SSRSelectorShow
	sse := ds.NewSSE(w, r)
	sse.MarshalAndMergeSignals(map[string]any{"SSRSelectorShow": store.SSRSelectorShow})
	sse.MergeFragmentTempl(SSRSelector(store.SSRSelectorShow), ds.WithoutViewTransitions())
}
```
This worked as expected when I used datastar v0.19.9. I explicitly turned off the view transitions using the `ds.WithoutViewTransitions()` option, but when I updated to  `github.com/starfederation/datastar v0.20.0`, and equivalent in the frontend the, the view transition was present. 

If I use `ds.WithViewTransitions()` instead, I get no transition. So it behaves the opposite right?

bencroker:
Looks like a duplicate of https://github.com/starfederation/datastar/pull/238.

bencroker:
Please review and merge, @delaneyj.

delaneyj (approved):
Already covered by #238, thank you and sorry about this!

PR #231, opened 2024-11-23
C# SDK

SpiralOSS:
I have at least one person that started poking at the C# side of the dotnet SDK (which is very broken in the `develop` branch). This should tide them over until I get some official unit test cases and expand the F# side of things.

For the majority of projects, the C# "surface" are these entities:
- builder.Services.AddDatastar<>()  // this injects the interfaces into ASP endpoints
- IDatastarSignalStore  // a user will implement this on their json DTO to represent the data-store
- IServerSentEventService  // this is the entry point to call merge|remove/Fragments|Signals methods

Most everything else is accessible but tucked away.


delaneyj (approved):
Thanks for all the effort!

PR #232, opened 2024-11-23
feat: Switch Docker-dev base image to Ubuntu, so headless chrome can work

khalwat:
Switch from Alpine to Ubuntu for the base image so that headless Chrome (and thus the tests) can work without shenanigans that seem worse:

https://stackoverflow.com/questions/70254649/rod-running-in-docker-alpine-get-error-chrome-linux-chrome-no-such-file-or-dir


delaneyj (approved):
:shipit:

PR #235, opened 2024-11-24
Support KV and boolean attributes in SDK

bencroker:
This change forces SDKs to parse attributes as an array of `key value` strings, _or_ an array of key-value pairs (and I’d even argue that only the latter should be allowed).

It also adds support for boolean attributes.

See the PHP implementation:
https://github.com/starfederation/datastar/blob/sdk/php/code/php/sdk/src/events/ExecuteScript.php#L50-L57


bencroker:
Agreed that this should be implemented as needed by each SDK, rather than become part of the spec.

PR #236, opened 2024-11-24
Support boolean attributes in execute script

bencroker:
This adds support for boolean attributes, so that `data: attributes disabled` is a valid data line.

See the related PR: https://github.com/starfederation/datastar/pull/235


delaneyj (approved):
This was a misunderstanding of the current behavior of `slice` with negative values.  THanks!

PR #237, opened 2024-11-25
Fix invalid http headers

ymijorski:
The http headers sent by Datastar are not valid at the moment. This results in some frameworks having trouble decoding the body of the request.

image


delaneyj (changes requested):
Change to use the `DATASTAR_REQUEST` const please
```ts
 "Content-Type": "application/json",
[DATASTAR_REQUEST]: "true",
```

delaneyj:
We talk on Discord, I'm gonna to add the content type as a constant at the SDK level

delaneyj:
Thanks for this, fixed in #241 as a stop gap.  We have to make changes in the SDK to unify this.

PR #238, opened 2024-11-25
update to swap with and without vt bools

zangster300:


bencroker:
Looks like a duplicate of https://github.com/starfederation/datastar/pull/230.

PR #251, opened 2024-11-26
Add platforms to taskfile

lllama:
Means that we download the correct version of tailwindcli for the current platform.


bencroker:
Thanks for the PR! Can you show a sample usage of it?

lllama:
So this is (/has been) fairly automatic. By adding the platforms then `task` will determine the platform it's on and only call tasks/cmds for that platform: https://taskfile.dev/usage/#platform-specific-tasks-and-commands

bencroker:
That’s what I was hoping. Tested locally and it works great, thanks!

PR #259, opened 2024-11-27
Delaneyj/issue258

delaneyj:
1. sseShared action wasn't merging properly
2. made the regex for token check more robust


delaneyj:
Closing for #266

PR #291, opened 2024-11-30
Delaneyj/issue216

delaneyj:


bencroker:
@delaneyj just a note that I see you reverted back to names that contain `store` like `marshalledStore` and `revisedStore`.

delaneyj:
closed in favor of #296

PR #297, opened 2024-12-01
Support explicit syntax

delaneyj:
Death to the $


library/src/plugins/official/backend/actions/sse.ts
```diff
@@ -0,0 +1,121 @@
...
+
+const isWrongContent = (err: any) => `${err}`.includes(`text/event-stream`);
+
+export const ServerSentEvents: ActionPlugin = {
+    type: PluginType.Action,
+    name: "sse",
+    fn: async (
+        { el: { id: elId }, signals },
+        url: string,
+        { method = "GET", headers = {}, onlyRemote = true }: {
```
bencroker: `method` should be converted to uppercase somewhere.

library/src/bundles/datastar.ts
bencroker, line 1: Reorder plugins.

bencroker:
Reminder to add the ability to filter `data-on-signals-changed` based on a signal path in dot notation.

Also, change the argument that `setAll()` and `toggleAll()` take from a regexp to a path in dot notation.

bencroker:
Reminder to switch to the following modifier syntax, to enable nestable signals in attribute keys.

```
data-signals-foo.bar?mod1_arg1_arg2&mod2_arg1_arg2=""
```

PR #312, opened 2024-12-06
Added Example output

rphumulock:


bencroker:
Made some tweaks and merged, thanks!

PR #313, opened 2024-12-06
Intellij Plugin

bencroker:
Adds autocomplete for Datastar to JetBrains IDEs.


bencroker:
Amazing work, thank you!!

PR #318, opened 2024-12-06
toggleAll & setAll use path instead of regex

### LINKED ISSUE #317: toggleAll & setAll use path instead of regex

delaneyj:
Fixes #317


bencroker (approved):

bencroker:
Great!

PR #336, opened 2024-12-09
Added alias to the commonly used paths

jmcudd:
Instead of `import { ActionPlugin, PluginType } from "../../../../engine/types";` you can now `import { ActionPlugin, PluginType } from "#engine/types";` and tsc will resolve the paths. This includes `#engine`, `#utils`, and `#vendored`.


library/tsconfig.json
```diff
@@ -17,7 +17,13 @@
     "noFallthroughCasesInSwitch": true,
     "baseUrl": "ts",
     "declaration": true,
-    "outDir": "dist"
+    "outDir": "dist",
+    "baseUrl": "./",
+    "paths": {
```
jmcudd, line 22: Additional mapping can be added here.

delaneyj (changes requested):
Like it in concept but this does not work with the `task build` bundling.

Rejecting until it does

PR #337, opened 2024-12-09
Added prettierrc to closely mimic current code

jmcudd:


bencroker:
Thanks for the PR but we’re avoiding NPM where possible. I’ll let @delaneyj make the final call on this.

delaneyj:
If you can make this work in VSCode and JetBrains without npm i'm on board.  Last I checked it uses the local node_modules prettier, and I'm not adding that evil in

jmcudd:
I've added a `.prettierrc` file to the project to make sure that any Prettier extensions, like the one I'm using with Vim (`prettier/vim-prettier`), don't automatically apply their default settings and mess up our existing code style. This won't add any npm modules to the project; it's just a simple configuration file to fix the issue I encountered where my Vim plugin was causing unnecessary formatting changes and cluttering up my commit history. The file is also supported by both VSCode and JetBrains IDEs, which means that if you have the Prettier extension in VSCode or if you're using any JetBrains IDE like IntelliJ IDEA or WebStorm, these editors will respect the `.prettierrc` settings. This helps ensure that everyone is using the same formatting rules, no matter which editor you prefer, and keeps the codebase consistent.

library/.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+    "tabWidth": 4,
+    "singleQuote": true,
+    "trailingComma": "all",
+    "bracketSpacing": true,
+    "arrowParens": "avoid",
+    "printWidth": 70,
+    "proseWrap": "never",
+    "arrowParens": "always"
```
bencroker, line 8: `arrowParens` has a duplicate with a different value.

delaneyj:
If it works without node_modules then there are some changes I'll probably make but do want use all on the same page

jmcudd:
You may want to run `npx prettier --write "**/*.{ts}"` in a single commit so all files inside `./library` conform to the prettierrc. I haven't familiarized myself enough with your current linting processes but it would be a good idea to incorporate it in a pre-commit process.

Can I ask why you're against using any node modules, its just a dev dependency and would not change the size of your bundles.

delaneyj:
Fixed with #361 , thanks for the push to clean this up!

PR #341, opened 2024-12-10
data-modifiers-*

### LINKED ISSUE #340: data-modifiers-*
after a long discussion on discord making modifiers into expressions.

delaneyj:
Fixes #340


library/src/plugins/official/dom/attributes/on.ts
```diff
@@ -59,72 +48,42 @@ export const On: AttributePlugin = {
...
         };
 
-        const debounceArgs = mods.get("debounce");
-        if (debounceArgs) {
-            const wait = argsMs(debounceArgs);
-            const leading = argsHas(debounceArgs, "leading", false);
-            const trailing = !argsHas(debounceArgs, "noTrail", false);
-            callback = debounce(callback, wait, leading, trailing);
+        if ("debounce" in mods) {
+            const time = mods?.debounce?.time || 100;
```
bencroker: @delaneyj we moved away from using `time` in previous version. Can we use `duration` here too?

PR #346, opened 2024-12-10
make modifiers spec compliant

### LINKED ISSUE #345: make modifiers spec compliant
https://validator.w3.org/nu/#textarea

```html
<!DOCTYPE html>
<html lang="">
<head>
<title>Test</title>
</head>
<body>
<p data-on-keydown:debounce_100ms:throttle_lead="value">Broke</p>
<p data-on-keydown_debounce.100ms_throttle.lead="value">Works</p>
</body>
</html>
```

delaneyj:
Fixes #345
 
Because we use `_` for local vs remote signals we have too many overloads.  Data-* attributes are very limited in valid names that don't resort to special hard to type characters... `__` seems to be the only valid option currently

```html
<p data-on-keydown__debounce_100ms__throttle_nolead="value">Works</p>
```


bencroker (approved):
If all tests are passing then this looks good to me!

PR #359, opened 2024-12-12
Module doesn’t export Datastar

### LINKED ISSUE #358: Module doesn’t export Datastar
Getting an error when trying to import Datastar (core and full) into a module.

```js
import {Datastar} from 'https://cdn.jsdelivr.net/gh/starfederation/datastar/bundles/datastar.js';
```

Results in:

```
Uncaught SyntaxError: The requested module 'https://cdn.jsdelivr.net/gh/starfederation/datastar/bundles/datastar.js' doesn't provide an export named: 'Datastar'
```

delaneyj:
Fixes #358 

Need to be explicit with ESM module build within esbuild call.  Also if you need to put on global here is an example.

```html
<!DOCTYPE html>
<html>
<head>
    <title>Home</title>
    <script type="module">
        import {Datastar} from './bundles/datastar.js'
        window.Datastar = Datastar
    </script>
</head>
<body data-signals-foo="1234">
    <h1>Home</h1>
    <p>Welcome to the home page.</p>
    <p data-text="foo.value"></p>
      <script async defer>
        window.addEventListener('DOMContentLoaded', () => {
            console.log(window.Datastar)
        })
    </script>
</body>
</html>
```


delaneyj:
Fixed #358

PR #370, opened 2024-12-13
Lowercase modifiers and arguments

### LINKED ISSUE #368: Lowercase modifiers and arguments
Modifiers and arguments should be converted to lowercase before comparing, to ensure there aren’t casing issues.

`noTrail` and `noLeading` should also become `notrail` and `noleading` in the library.

delaneyj:
Fixes #368


library/src/utils/tags.ts
```diff
@@ -13,7 +13,7 @@ export function argsMs(args: Set<string>) {
   return 0
 }
 
-export function argsHas(args: Set<string>, arg: string, defaultValue = false) {
-  if (!args) return defaultValue
-  return args.has(arg)
+export function tagHas(tags: Set<string>, tag: string, defaultValue = false) {
+  if (!tags) return defaultValue
+  return tags.has(tag.toLowerCase())
```
bencroker, line 18: @delaneyj this `toLowerCase()` seems unnecessary, since it refers to internal checks only.

delaneyj: for now... what if it's an external plugin?

bencroker: Yeah, good point.

PR #376, opened 2024-12-14
Re-add manual building instructions

Superpat:
Useful for building with docker as well since it gives indications about the main task commands.


bencroker:
This looks good so far. Git-LFS is no longer required, can you please remove?

Superpat:
Oups, thought I had removed all mentions of that. It's done !

bencroker:
Merged, thanks!

Issue #383, opened 2024-12-16
Bundler is broken


zangster300:
pathing issue, fix incoming...

PR #385, opened 2024-12-16
fix for issue 383

zangster300:
adds an alias map to the `esbuild` config to resolve lib files to use the tmp directory to find files

#383


bencroker:
That fixed it, thanks!

Issue #382, opened 2024-12-16
Support infinite retries, make initial retry time configurable


magiconair:
For some use cases I would need an indefinite retry since otherwise we would have to reload the page to re-establish the connection. As a followup to #190 I would like to suggest the following additions:

* Document that the initial retry wait time is `1s`
* Provide an `retryInitialWaitMs` and set it to `1s`
* Use `retryMaxCount: -1` as "retry indefinitely"
* Use `retryMaxCount: -2` as "do not retry"

delaneyj:
`retryMaxCount:Number.MAX_SAFE_INTEGER` is in practice infinite
`retryMaxCount:0` is do not retry

PR #388, opened 2024-12-17
reference/sse: clarify retry behavior

magiconair:
* document initial wait time
* provide an example for "infinite" retries

### DIFF
--- a/site/static/md/reference/action_plugins.md
+++ b/site/static/md/reference/action_plugins.md
@@ -26,9 +26,9 @@ The `sse()` action takes a second argument of options.
 - `includeLocal` - Whether to include local signals (those beggining with an underscore) in the request (defaults to `false`).
 - `headers` - An object containing headers to send with the request.
 - `openWhenHidden` - Whether to keep the connection open with the page is hidden (defaults to `false`). Useful for dashboards but can cause a drain on battery life and other resources.
-- `retryScaler` - A number that scales the retry time (defaults to `2`).
+- `retryScaler` - A number that scales the retry time (defaults to `2`). The initial wait time is `1s`.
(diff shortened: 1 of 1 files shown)


bencroker:
> The initial wait time is `1s`.

I find this comment confusing rather than clarifying. The initial wait time is `retryScaler` multiplied by `retryMaxWaitMs`, which happens to be 1 second given their default values, which are documented.

> Use a large number (e.g. `Number.MAX_SAFE_INTEGER`) for "infinite" retries.

Does this really need to be stated? Definitely not “infinite”.

magiconair:
I was looking in the documentation for these two values: What does the backoff algorithm start with and how can I turn off the retry limit since I want a constant, predictable, infinite retry for the '/event' endpoint for screens where there is no user to hit F5. I cannot afford the screen to stall.

Initial retry time: Is this better? The default is not on the page. Probably in the code only.

As for disabling the retry limit: I think a sentinel value or a separate flag to disable the limit makes the intent more clear (and maybe less error-prone than `data-on-load="sse('/events', {retryScaler:1,retryMaxCount:Number.MAX_SAFE_INTEGER})"`. #382 says that using `Number.MAX_SAFE_INTEGER` achieves virtually the same thing. It is logical but not obvious - at least not to me.

bencroker:
I know you mean well but I find the contributions misleading more than providing any real value. The “initial wait time” is not 1 second! It is a function of two settings.

bencroker:
So after looking into this a bit more, I realised that I misunderstood, and that the retry interval is hard-coded to be 1 second. I’ve created a PR to make this configurable in https://github.com/starfederation/datastar/pull/393. Apologies for my misunderstanding.

PR #391, opened 2024-12-17
Added JS SDK implementation.

jmcudd:


delaneyj (changes requested):
Looking great, we'd like to see this use the consts provided by the build process.  Check out the other SDKs for details

Superpat:
@jmcudd I did not see this pop up, but I already had a working js / ts sdk on the old repo, just never had time to port it to this one. The code is in #402 if you want to take a look at it.

I think I'll take the doc comments from your code if you dont mind, I forgot to add some to mine. I also went with separate implementations of the sdk for different runtimes, since deno, bun and node dont do everything the same way, but I'll have to actually test that, maybe they all use the same Request / Response type.

jmcudd:
I'm having some trouble getting the consts template to render for some reason.

jmcudd:
Also I think I need to add a build pipeline to publish the npm module.

Superpat:
> I'm having some trouble getting the consts template to render for some reason.

I you want to keep building a parallel sdk to mine I recommend looking at the ts const template in my branch. That one renders correctly.

delaneyj:
I think we need to coalesce on one.  @Superpat's version is definitely farther ahead atm.

PR #392, opened 2024-12-17
Replace http.Flusher with http.ResponseController for SSE in Go SDK 

drahtzieher:
Using ResponseController instead of Flusher is more appropriate since Go 1.20:
https://tip.golang.org/doc/go1.20#http_responsecontroller
https://www.alexedwards.net/blog/how-to-use-the-http-responsecontroller-type

It changes the ServerSentEventGenerator struct as seen below, replacing the flusher with a ResponseController.


delaneyj (changes requested):
Please add back in the comments about panic being a deliberate choice

drahtzieher:
I have re-added it to the panic when we flush the headers.

delaneyj (approved):

PR #394, opened 2024-12-17
Fix & unify CDN in docs/examples

dalito:
The Python examples were not working due to the wrong CDN URL.

I made all URLs the same: `https://cdn.jsdelivr.net/gh/starfederation/datastar/bundles/datastar.js`.

Let me know if you prefer: `https://cdn.jsdelivr.net/gh/starfederation/datastar@develop/bundles/datastar.js` which was used in the introduction (but only there)

### DIFF
--- a/site/routes_home.templ
+++ b/site/routes_home.templ
@@ -9,7 +9,7 @@ import (
 
 templ Home() {
 	{{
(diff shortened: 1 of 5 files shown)


bencroker:
Merged, thanks!

PR #397, opened 2024-12-18
ReadSignalsWithLimit added, no API breakage

iansmith:
I wanted to limit the size of uploads because file uploads of unlimited size are an attack vector.  I also wanted to control the size from user code because I need to allow users to upload several large images at once.

`ReadSignalsWithLimit` is now the implementation of `ReadSignals` where it uses a default max size.  I also added and error that folks can check for if they want to detect this situation in user code.

### DIFF
--- a/sdk/go/signals.go
+++ b/sdk/go/signals.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


delaneyj (changes requested):
Like the idea, needs to be changed to an options API, something like

```go
 ReadSignals(r *http.Request, signals any, opts ...ReadOption) error
 ```

bencroker:
@iansmith any update on this?

delaneyj:
?

delaneyj:
Please reopen when you make the changes per request above

PR #398, opened 2024-12-18
attempt to make the 3 signals of file upload get set

iansmith:
I have no idea if this actually is the correct way to do this or, indeed, if I understand the problem at all.  I don't really know TS well enough to be doing this type of thing.

I found that the file upload wasn't setting all three signals--[blah]Names ad [blah]Mimes were not present when I called the data upload endpoint.


bencroker:
This is not at all correct. Please create a GitHub issue first, describing the problem, so we can help troubleshoot and figure out if this is a bug or a misunderstanding on your part.

bencroker:
Please start by creating an issue with a simple test case so we can troubleshoot and see if this is indeed a bug. If it is, then it impacts the more serious issue of upserting of signals, rather than simply file uploads.

Issue #293, opened 2024-11-30
Figure out forms


bencroker:
Figure out working with forms in Datastar (aka Delboy’s Magnum Opus).

For reference: https://github.com/starfederation/datastar/discussions/82, https://github.com/starfederation/datastar/discussions/136

bencroker:
For reference: https://github.com/starfederation/datastar/discussions/82, https://github.com/starfederation/datastar/discussions/136

Superpat:
First of all, do we want to figure out forms before v1.0 ? 

Second of all, how much magic do we want ?

- Should the name attribute become the name of a signal ? 
- Should it be a nested signal under the form's namespace  ? 
- Should the native browser validation be run on inputs before their value is propagated to their bound signal ?
- Should changes to a signal be propagated to it's bound input if the value does not pass that input's validation ?

I'm sure there's other stuff, but that's the questions I had.

dkotik:
> First of all, do we want to figure out forms before v1.0?

Absolutely!

> Should it be a nested signal under the form's namespace?

Certainly!

I wondered if there is an ability to scope $post body here: https://github.com/starfederation/datastar/discussions/302 I think this issue is entirely related. Right now there is too much magic. JSON encoding a step away from hypermedia. Should it perhaps be an extension, with form data being the default? Seem sensible to me.

Also, using `<input>` and such without `<form>` tag is technically incorrectly formed HTML - it gets picked on by linters. I suggest that data submission should be handled via forms, not the global store of everything.

Superpat:
> > First of all, do we want to figure out forms before v1.0?
> 
> Absolutely!

I ask the question because it could very well be a separate plugin that does not affect the datastar core, so released on a separate occasion. Though stuff like validation of data-bind would occasion changes to other plugins.


> I wondered if there is an ability to scope $post body here: #302 I think this issue is entirely related. Right now there is too much magic. JSON encoding a step away from hypermedia. Should it perhaps be an extension, with form data being the default? Seem sensible to me.
> 
> Also, using `<input>` and such without `<form>` tag is technically incorrectly formed HTML - it gets picked on by linters. I suggest that data submission should be handled via forms, not the global store of everything.

I highly doubt @delaneyj is going to want to change the default format of post requests data in datastar, but the ability to use a different encoding for the data could be very useful. I know I've had a datastar project that used an external api that only accepted  formdata encoded images and having to reparse the file as formdata was annoying.

bencroker:
[...]

PR #400, opened 2024-12-18
Add `form` option to `sse()` action

bencroker:
This adds a new `contentType` option to the `sse()` action as discussed in https://github.com/starfederation/datastar/issues/293.

```html
<form>
  <input name="foo">
  <button data-on-click="sse('/endpoint', {contentType: 'form'})">
    Submit GET request
  </button>
  <button data-on-click="sse('/endpoint', {contentType: 'form', method: 'post'})">
    Submit POST request
  </button>
</form>

<button data-on-click="sse('/endpoint', {contentType: 'form', selector: '#myform'})">
  Submit GET request from outside the form
</button>
```

When the `contentType` option is set to `form`, the following steps are taken.

1. If a `GET` request, set `Content-Type` to `application/x-www-form-urlencoded`, otherwise `multipart/form-data`.
2. Find the closest form element or use the `selector`, if one is provided. If no form is found, throw an error.
3. Add an event listener to prevent submission of the form.
4. Perform form validation. If invalid, report the validity, remove the event listener, and return.
5. If a `GET` request, set the query params, otherwise the request body.
6. Remove the event listener.


site/routes_examples_form_data.go
```diff
@@ -0,0 +1,52 @@
...
+			return
+		}
+
+		sse.ExecuteScript(fmt.Sprintf(`alert('Form data received via GET request: %s')`, jsonData))
+	})
+
+	examplesRouter.Post("/form_data/data", func(w http.ResponseWriter, r *http.Request) {
+		sse := datastar.NewSSE(w, r)
+		
+		err := r.ParseMultipartForm(1 << 20)
```
bencroker: @delaneyj please help me figure out why parsing the multipart form for `POST` requests is failing in the Go example.

delaneyj: You have to parse forms before setting up SSE, i'm all backends

bencroker: Thanks!

bencroker: @delaneyj I’m still getting `Failed to parse multipart form`.

Superpat:
Does this work if you use on-submit instead of on-click ?

library/src/plugins/official/backend/actions/sse.ts
```diff
@@ -30,6 +31,8 @@ const isWrongContent = (err: any) => `${err}`.includes(`text/event-stream`)
 
 export type SSEArgs = {
```
Superpat, line 31: The SSEARGS type could be made stricter. Here's an example:

```
export type SSEArgs = {
    method: METHOD
    contentType: 'form'
    selector?: string
    headers?: Record<string, string>
    openWhenHidden?: boolean
} | {
    method: METHOD
    contentType: 'json'
    headers?: Record<string, string>
    openWhenHidden?: boolean
    includeLocal?: boolean
}
```
This way we encode that includeLocal can only be used with contentType: 'json' and selector can only be used with contentType: 'form', since unless I'm mistaken, form submit dont include the signals in any case and selector makes no sense if it's not a form.

library/src/plugins/official/backend/actions/sse.ts
```diff
@@ -138,15 +146,47 @@ export const SSE: ActionPlugin = {
...
+        const formEl = selector ? document.querySelector(selector) : el.closest('form');
+        if (formEl === null) {
+          if (selector) {
+            throw dsErr('SseFormNotFound', { selector })
+          } else {
+            throw dsErr('SseClosestFormNotFound')
+          }
+        }
+        const preventSubmit = (evt: Event) => evt.preventDefault()
+        formEl.addEventListener('submit', preventSubmit)
```
Superpat: If somebody calls `sse` from a data-on-submit, I dont think this will prevent it's default behaviour. This would probably come as a surprise to anybody trying to use data-on-submit.

library/src/plugins/official/backend/actions/sse.ts
bencroker: Good point, but we may want to use `selector` for other content types at some point. I feel like this suggestion, while valid, adds unnecessary strictness. Any thoughts, @delaneyj ?

library/src/plugins/official/backend/actions/sse.ts
bencroker: You’re right. Is there anyway around this, besides documenting that `data-on-submit__prevent` must be used?

library/src/plugins/official/backend/actions/sse.ts
Superpat: I get what you mean, it's easy to go too far with types. But I'd consider this relatively tame still. Until we do add a content type that can use selector as well, this does makes it clear to anybody reading the code or extending it. 

We can make it more DRY by keeping headers, openWhenHidden and method in a separate interface:

```
{
    method: METHOD
    headers?: Record<string, string>
    openWhenHidden?: boolean
} & ({ 
    contentType: 'form'
    selector?: string
} | {
    contentType: 'json'
    includeLocal?: boolean
}) 
```

delaneyj: > Good point, but we may want to use `selector` for other content types at some point. I feel like this suggestion, while valid, adds unnecessary strictness. Any thoughts, @delaneyj ?

This type constraint is correct. It doesn't mean we can't change it to be more open later. It's a good idea

library/src/plugins/official/backend/actions/sse.ts
delaneyj: maybe always prevent if it onSubmit?  If you are using Datastar, you are always preventing default form stuff

library/src/plugins/official/backend/actions/sse.ts
bencroker: Added.

library/src/plugins/official/backend/actions/sse.ts
bencroker: Ok, `data-on-submit` attributes now have `preventDefault()` applied to the event by default.

bencroker:
@delaneyj there’s a TODO in `sse.ts` that needs your input, please.

delaneyj:
Looks like we didn't need to do any engine.ts changes, just cleanup the form event listener. @bencroker

eltimn:
I was working on a form yesterday and was thinking about this last night. I believe that forms should always be submitted by listening for the `submit` event on the form tag instead of using a click handler on a button. The main reason is that it usually breaks being able to hit the return key while inside the form to submit it. My preference would be to have a `data-on-submit` attribute that would go on the form tag itself, so it will work no matter how the form was submitted.

Great framework, btw. I started working on a project with it a few days ago and am liking it a lot.

eltimn:
For reference, since I already had it open: [HTMLFormElement: submit event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event)

delaneyj:
> I was working on a form yesterday and was thinking about this last night. I believe that forms should always be submitted by listening for the `submit` event on the form tag instead of using a click handler on a button. The main reason is that it usually breaks being able to hit the return key while inside the form to submit it. My preference would be to have a `data-on-submit` attribute that would go on the form tag itself, so it will work no matter how the form was submitted.
> 
> Great framework, btw. I started working on a project with it a few days ago and am liking it a lot.

Both should work actually to make up for forms being the worst part of HTML

delaneyj (approved):

Superpat:
> I was working on a form yesterday and was thinking about this last night. I believe that forms should always be submitted by listening for the `submit` event on the form tag instead of using a click handler on a button. The main reason is that it usually breaks being able to hit the return key while inside the form to submit it. My preference would be to have a `data-on-submit` attribute that would go on the form tag itself, so it will work no matter how the form was submitted.

@eltimn @bencroker made sure `data-on-submit` would work.

PR #402, opened 2024-12-18
Ts sdk

Superpat:
New pull request for typescript / javascript sdk. This is an abstract sdk, it's serverSentEventGenerator needs to be 
extended by a concrete runtime. But all tests pass. 

My next step is the node sdk, which I'd already written, so I just need to port to this new version, I'll push it to this branch later, but at least people can review the abstract class.


delaneyj (changes requested):
I'd like to see a README with basic usage, otherwise I think this is great,  @bencroker ?

bencroker:
Agreed, looks great!

Superpat:
I've added a node impl, but I'm trying to make it cross-runtime. The readSignal works fine, but I havent yet found out the best method for the streaming. I need to read up on WritableStreams.

Superpat:
So I think I've managed to make it cross-runtime (it should even work in service workers). Had to make some changes to the sdk interface because of how js uses callbacks for everything. Now you need to call the static `stream` method on the class and pass it a function. The function will have access to the class and can do all the sdk methods. But I'm open to suggestions on how to do it better.

Now I'm struggling with getting vite to start a node server and I think my next step will be replacing vite with esbuild so the `task build` task can build the sdk at the same time.

delaneyj:
Do you want us to approve where it's at, or wait for your esbuild changes?

Superpat:
No it can wait, I just pushed and gave an update to keep people in the loop. I'll call you if I struggle with esbuild.

Superpat:
@delaneyj can you review the changes to the build code ? I'm sure you have notes. Otherwise everything works on both deno and node. I'm writting docs next then we can merge. 

I want to do a command line based test suite for sdks, but that's orthogonal.

Superpat:
Dont review the code too much yet, I'm still going through some minor fixes as I realize I missed a few things in the spec for events other then `datastar-merge-fragments`.

nickchomey:
For whatever it's worth, I'm particularly interested in trying this out in a service worker (it was mentioned above that this should be possible). 

Essentially, I'd like to send a bunch of data and templates to the browser, store in indexeddb, and then allow the service worker to function as the backend for datastar (eg if the user is offline, or just to minimize latency, bandwidth usage, backend server power etc...).

bencroker:
@Superpat is this ready for review or does it need more work? If the latter, can you put the PR in draft status, otherwise I get a notification to review it for each individual commit.

Superpat:
@delaneyj @bencroker @jmcudd I do think this is ready for an in depth review.

jmcudd:
Looks good to me. I ended up iterating a bit more on my [own implementation](https://github.com/jmcudd/datastar-ssegen/blob/main/index.js), I've been using it for the past couple weeks in my own projects. I didn't test it with Deno, but it works with Node, Bun, and Elysia. Feel free to borrow any ideas I had. 

Is there a way to access the headers for the response? I found that some frameworks like Elysia set headers differently, so I just made the headers available on the ServerSentEventGenerator object.

You might want to look at [Elysia's use](https://elysiajs.com/essential/handler.html#stream) of `yield` for handling streams, not sure how that would work with your implementation.

Superpat:
@jmcudd thanks, I see you strip out comment blocks in executeScript. Magiconair's old implementation also minimised the html fragments. I'm not convinced this is necessary yet, as it does complexify those functions a bit and I dont think you're encouraged to send much js code. It's mostly intended as an escape hatch to use browser api's. 

For a real js script you'd be better off just including it as a link, since then the browser can handle caching and the likes.

I havent used Elysia or Bun yet, but Bun seems to aim for much more node compatibility then node does, while still supporting web standards as much as deno does. From looking at it's front page, I think a bun datastar app could use both the node and web versions. But I'll try to setup a working bun example.

Superpat:
> For whatever it's worth, I'm particularly interested in trying this out in a service worker (it was mentioned above that this should be possible).

I'd be open to a working example of the sdk being used in a service worker, but I wont be testing that myself until I have a concrete need for it. All I know is that all the api's I used should be web standard.

nickchomey:
> > For whatever it's worth, I'm particularly interested in trying this out in a service worker (it was mentioned above that this should be possible).
> 
> I'd be open to a working example of the sdk being used in a service worker, but I wont be testing that myself until I have a concrete need for it. All I know is that all the api's I used should be web standard.

Now that this SDK is "ready", I'll see if I can figure out how to get it working in a service worker tomorrow. 

> You might want to look at [Elysia's use](https://elysiajs.com/essential/handler.html#stream) of yield for handling streams, not sure how that would work with your implementation.

This is something I had in mind as well - you could start streaming a response as it is being rendered, rather than wait for the whole thing to be ready. Something like this was implemented in the [Service Worker Templating Language](https://thepassle.netlify.app/blog/service-worker-templating-language-(swtl)) library. Here's the [relevant code](https://github.com/thepassle/swtl/blob/ec5b02a9383a3c722ea9beb079dcfbe936e34a2c/packages/core/render.js#L67) and the [npm package](https://www.npmjs.com/package/swtl#iterablesresponses)

jmcudd:
> For a real js script you'd be better off just including it as a link, since then the browser can handle caching and the likes.

So far the most useful code execution use-case for me has been dynamic imports which are a bit easier to manage than big code blocks that have to be minified at runtime. Generally you wouldn't do this, but it's great for dynamically loading in large web components when they are needed rather than on first load. Not sure about if browsers cache dynamic imports, I imagine not, but who knows. 

Something like this. 

```javascript 
yield sse.ExecuteScript(`import('./my-web-componet.js');`);
//followed by
yield sse.MergeFragment(`<div id="target"><my-web-component></my-web-component></div>`)
```

Superpat:
> > You might want to look at [Elysia's use](https://elysiajs.com/essential/handler.html#stream) of yield for handling streams, not sure how that would work with your implementation.

> This is something I had in mind as well - you could start streaming a response as it is being rendered, rather than wait for the whole thing to be ready. Something like this was implemented in the [Service Worker Templating Language](https://thepassle.netlify.app/blog/service-worker-templating-language-(swtl)) library. Here's the [relevant code](https://github.com/thepassle/swtl/blob/ec5b02a9383a3c722ea9beb079dcfbe936e34a2c/packages/core/render.js#L67) and the [npm package](https://www.npmjs.com/package/swtl#iterablesresponses)

So my understanding is that would make the transfers to the client faster, though I think the events would need to be quite large for that to be really noticeable no ?

I hesitate to add any kind of parsing like that since I feel it would be easy to get something wrong, but I'm not fundamentally against it. Maybe it could wait until this pull request is merged and be the subject of another pull request ?

EDIT: For it to be truly streaming as it renders, the various events would need to expose their stream so that a user could pipe in their content as it's being calculated. I think there might be a way to do that for the {fragment|data|script} parts of the events without changing the rest of the code too much. But I still think it's a premature optimization, especially since it would change the api quite a bit.

nickchomey:
Fair enough to leave it til later if you think it won't be easy to do! 

You're probably right that it won't be necessary in general, perhaps especially since Delaney has shown some pretty high frequency/large fragment swap use cases working well without something like this. 

Still, such streaming is part of the http and web api for a reason. I'll open an issue for it where it can be discussed. 


I'll try to provide some service worker feedback today.

Superpat:
Yeah, the crazy stuff @delaneyj does performance wise is why I'm hesitating to complexify the code. I suspect if you're merging in something that is that huge you can probably break it up in different events.

delaneyj:
I've noticed at least in my past that Go in real world applications tends to be about 6 to 12 times faster than TS based ones.  Do whatever you think is right for performance, this is all backend code so it doesn't really matter to me what the complexity needs to be to do the right thing. I trust y'all to make good choices

Superpat:
I'd vote for keeping it simple until someone can demonstrate that they need a faster method. 

My previous streaming datastar app had no performance problems with html rendering.

nickchomey:
The other thing is that it seems like if responses were being streamed, it might mess with idiomorph. To avoid that, you'd probably have to manually delineate the chunks and have each one target more specific Dom elements and use append/prepend etc..., rather than send and merge them all within a parent element. I could be wrong though.

nickchomey:
Ive tested this in a service worker - i modified the deno test app to use Hono and then put the routing etc in a shared-router module that the deno and service-worker scripts both import. They're similar enough to each other, but with some minor differences. Hono makes it easy though. 

Anyway, its not well-implemented and I only did a simple fragment merge, but it *seems* to work. 

https://github.com/user-attachments/assets/ae219feb-e483-4a64-a954-97d662dfe47b


However, I do notice what seems to be an issue: the deno and node examples are using `sse()` rather than `$get()` from v1 beta. Given that that's the only D* tag in use, is that the only instance of something being out of date (i think the rest of the SDK really just handles SSE creation and is agnostic to the actual content of the html that will drive D* in the browser)?

Also, it is also pulling in v0.21.4 from jsdeliver rather than 1.0.0-beta.1. Perhaps the beta in the tag makes it not default to that version?

nickchomey:
Deno's linter shows these errors - no idea if they matter or not. Basically all just using `any` and `async` without `await`

```
deno lint abstractServerSentEventGenerator.spec.ts abstractServerSentEventGenerator.ts consts.ts types.ts vite-env.d.ts web/deno.ts node/node.ts node/serverSentEventGenerator.ts web/serverSentEventGenerator.ts 
error[no-explicit-any]: `any` type is not allowed
  --> /ts/src/types.ts:54:47
   | 
54 |     [DatastarDatalineSignals]: Record<string, any>;
   |                                               ^^^
   = hint: Use a specific type other than `any`

  docs: https://lint.deno.land/rules/no-explicit-any


error[no-explicit-any]: `any` type is not allowed
  --> /ts/src/abstractServerSentEventGenerator.ts:62:59
   | 
62 |     private eachOptionIsADataLine(options: Record<string, any>): string[] {
   |                                                           ^^^
   = hint: Use a specific type other than `any`

  docs: https://lint.deno.land/rules/no-explicit-any


error[no-explicit-any]: `any` type is not allowed
   --> /ts/src/abstractServerSentEventGenerator.ts:102:46
    | 
102 |     public mergeSignals(data: Record<string, any>, options?: MergeSignalsOptions): ReturnType<typeof this.send> {
    |                                              ^^^
    = hint: Use a specific type other than `any`

  docs: https://lint.deno.land/rules/no-explicit-any


warn: await isn't allowed in non-async function at file:///ts/src/node/serverSentEventGenerator.ts:43:15

          await streamFunc(new ServerSentEventGenerator(req, res));
                ~~~~~~~~~~
error[no-await-in-sync-fn]: Unexpected `await` inside a non-async function.
  --> /ts/src/node/serverSentEventGenerator.ts:43:9
   | 
43 |         await streamFunc(new ServerSentEventGenerator(req, res));
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = hint: Remove `await` in the function body or change the function to an async function.

  docs: https://lint.deno.land/rules/no-await-in-sync-fn


error[no-process-globals]: NodeJS process global is discouraged in Deno
  --> /ts/src/node/serverSentEventGenerator.ts:73:43
   | 
73 |             const url = new URL(`http://${process.env.HOST ?? 'localhost'}${request.url}`);
   |                                           ^^^^^^^
   = hint: Add `import process from "node:process";`

  docs: https://lint.deno.land/rules/no-process-globals


error[require-await]: Async arrow function has no 'await' expression or 'await using' declaration.
  --> /ts/src/node/node.ts:15:51
   | 
15 |         ServerSentEventGenerator.stream(req, res, async (stream) => {
   |                                                   ^^^^^
   = hint: Remove 'async' keyword from the function or use 'await' expression or 'await using' declaration inside.

  docs: https://lint.deno.land/rules/require-await


error[require-await]: Async arrow function has no 'await' expression or 'await using' declaration.
  --> /ts/src/node/node.ts:23:59
   | 
23 |                 ServerSentEventGenerator.stream(req, res, async (stream) => {
   |                                                           ^^^^^
   = hint: Remove 'async' keyword from the function or use 'await' expression or 'await using' declaration inside.

  docs: https://lint.deno.land/rules/require-await


error[no-explicit-any]: `any` type is not allowed
  --> /ts/src/node/node.ts:74:67
   | 
74 |                     stream.mergeSignals(signals as Record<string, any>, options || undefined);
   |                                                                   ^^^
   = hint: Use a specific type other than `any`

  docs: https://lint.deno.land/rules/no-explicit-any


error[no-explicit-any]: `any` type is not allowed
 --> /ts/src/vite-env.d.ts:9:27
  | 
9 |   interface Assertion<T = any> extends CustomMatchers<T> {}
  |                           ^^^
  = hint: Use a specific type other than `any`

  docs: https://lint.deno.land/rules/no-explicit-any


error[no-explicit-any]: `any` type is not allowed
  --> /ts/src/abstractServerSentEventGenerator.spec.ts:46:27
   | 
46 |     event: Record<string, any>,
   |                           ^^^
   = hint: Use a specific type other than `any`

  docs: https://lint.deno.land/rules/no-explicit-any
  ```

Superpat:
> Ive tested this in a service worker - i modified the deno test app to use Hono and then put the routing etc in a shared-router module that the deno and service-worker scripts both import. They're similar enough to each other, but with some minor differences. Hono makes it easy though.
> 
> Anyway, its not well-implemented and I only did a simple fragment merge, but it _seems_ to work.

Great !


> However, I do notice what seems to be an issue: the deno and node examples are using `sse()` rather than `$get()` from v1 beta. Given that that's the only D* tag in use, is that the only instance of something being out of date (i think the rest of the SDK really just handles SSE creation and is agnostic to the actual content of the html that will drive D* in the browser)?
> 
> Also, it is also pulling in v0.21.4 from jsdeliver rather than 1.0.0-beta.1. Perhaps the beta in the tag makes it not default to that version?

Oh good catch, I forgot to update it. It has nothing to do with the sdk though, it's just the example that needs to be updated.

Superpat:
> Deno's linter shows these errors - no idea if they matter or not. Basically all just using `any` and `async` without `await`

Yeah, I'm not sure why the warnings I get from the ts lsp are different from the ones by deno lint. 

Though I dont know why it complains about the async without await, since ts complains that I'm not returning a promise if I dont do that.

bencroker:
Where are we with this, @Superpat?

Superpat:
Sorry It took me a longer time then I thought to fix the issues @nickchomey brought up, mostly because node and deno dont agree on wether imports should have extensions.. 

Ultimately I went with using deno dnt to build the node packages, it's not the perfect solution since it requires deno to be installed if you want to build the library for node, but it's the easiest solution I could find and it works. Also deno is a more pleasant env to dev in anyways, since I dont need to install any other tools to get linting et all.

Superpat:
The only remaining issue is http2 support. In my mind, node http was already http2 with fallback to http1, but I'm realising now that it's strictly http1 and you need to use node:http2 if you want http2. I'll figure that out another day, but at least we dont have to worry about safari not accepting the keepalive header :P

Superpat:
@delaneyj @bencroker To be clear, I think this should be merged now and http2 support can be added later, since it's a different api then http1.

I havent updated any of the website documentation with info about the ts sdk yet, but just let me know if I should before the merge.

nickchomey:
I think http2 support would be important since d* may very well hold at least 1 connection open and http1 only allows for 6 connections.

Superpat:
I'm not saying I wont add it, but @delaneyj asked me if we could merge and at this point I would say yes. The 6 connexion limit is browser based, not server based, so I think you could easily use a proxy to support http2.

Superpat:
I'd also like the shell based test suite to be available for use asap. Since it would be useful for other sdk devs.

bencroker:
Merged, thanks!!

PR #406, opened 2024-12-19
Fix bundler not exporting module

### LINKED ISSUE #403: Packages created by https://data-star.dev/bundler can't be imported as modules
The documentation for the [Javascript API](https://data-star.dev/reference/javascript_api) recommends for edge cases to import Datastar as a module like the following:

```html
<script type="module">
    import { Datastar } from 'datastar.js'
    
    Datastar.apply(document.body)
</script>
```

If I create a package with the [Bundler](https://data-star.dev/bundler) then the `datastar.js` is created **without** the module syntax.

I think the reason for the missing module syntax is the missing option for `Format: api.FormatESModule` here:

https://github.com/starfederation/datastar/blob/8c5edc7ee73b1a72f10ec558825776090d8d0fee/site/routes_bundler.go#L311-L323

The release process creates a `datastar.js` file too and copies it to https://github.com/starfederation/datastar/blob/main/bundles/datastar.js and this file has the module syntax, because the option for `Format` is set:

https://github.com/starfederation/datastar/blob/8c5edc7ee73b1a72f10ec558825776090d8d0fee/build/run.go#L61-L76

bencroker:
This adds the version comment using `Banner` and sets the `Format` to `api.FormatESModule`. It still doesn’t resolve #403 though, as the module is not exported. Can you please take a look, @delaneyj?


stefanohrmann:
@bencroker I think you have to adjust the [template of intermediate datastar.ts](https://github.com/starfederation/datastar/blob/bugfix/bundler/site/routes_bundler.qtpl)

```txt
{% func bundlerContent(manifest PluginManifest) %}

import { Datastar as DS } from "../engine";
{%- for _, p := range manifest.Plugins -%}
import { {%s p.Name %} } from "{%s p.Path %}";
{%- endfor -%}

DS.load(
{%- for _, p := range manifest.Plugins -%}
    {%s p.Name %},
{%- endfor -%}
);

export const Datastar = DS;
{% endfunc %}
```

For reference I used the bundled [datastar.ts](https://github.com/starfederation/datastar/blob/develop/library/src/bundles/datastar.ts) used by the release process.

bencroker:
Good catch, @stefanohrmann!

I’ve fixed and this is ready for @delaneyj to review.

delaneyj (approved):

PR #412, opened 2024-12-21
Dispatch an SSE error event

bencroker:
Dispatches an error event for responses with status codes `>= 400`. The behavior stays the same, but it allows 4xx and 5xx errors to be detected.


library/src/vendored/fetch-event-source.ts
```diff
@@ -363,13 +363,7 @@ export function fetchEventSource(
...
               dispose()
               // Max retries hit, check your server or network connection
-              reject(
-                dsErr('SSE_MAX_RETRIES', {
-                  retryInterval,
-                  retryMaxCount,
-                  ...rest,
-                }),
-              )
+              reject()
```
delaneyj: I don't grok why the max retries was removed

delaneyj (approved):

PR #416, opened 2024-12-22
Fix first quiz example

datashaman:
In the first example, there is no `answer2` signal, resulting in an error on the console.

Screenshot 2024-12-22 at 08 42 29


bencroker:
Merged, thanks!

PR #429, opened 2024-12-27
Can `.value` go away?

delaneyj:


bencroker:
Reminder to strip `$` off signal definitions:
`data-signals-$foo="1"`, `data-signals="{$foo: 1}"`, `data-bind-$foo`, ``data-bind="$foo"`

Issue #426, opened 2024-12-25
Empty `filesMimes` and `filesNames`


iansmith:
@bencroker As you requested, this is a simple test case to show that I don't understand what is going on with the other two values that I expected to get from the client.  You'll notice that the json body doesn't appear to have the other two fields on it when it prints them out.

@delaneyj I'm probably overcomplicating it.  

```

package main

import (
    "io"
    "log"
    "net/http"
    "strings"

    "[github.com/go-chi/](http://github.com/go-chi/)chi/v5"
    datastar "[github.com/starfederation/datastar/sdk/go](http://github.com/starfederation/datastar/sdk/go)"
)

const page = `
<html>
<head>
    <script type="module" src="https://cdn.jsdelivr.net/gh/starfederation/datastar/bundles/datastar.js"></script>
    <script type="module">


    </script>
</head>

<body>
    <div data-signals="{'files': [], 'filesMimes':[], 'filesNames':[]}">
        <input
            type="file"
            data-bind="files"
            multiple
        />  
    <button
      data-on-click="sse('/upload', {method: 'post'})">
      Submit
    </button>
    </div>

    <div >
        <span data-text="ctx.signals.JSON()"></span>
    </div>

</body>
</html>
`

type Holder struct {
    Files []string `json:"files"`
    Names []string `json:"filesNames"`
    Mimes []string `json:"filesMimes"`
}

func main() {
    router := chi.NewRouter()
    router.Get("/*", func(w http.ResponseWriter, r *http.Request) {
        w.Header().Add("Content-Type", "text/html")
        io.Copy(w, strings.NewReader(page))

    })
    router.Post("/upload", func(w http.ResponseWriter, r *http.Request) {
        err := r.ParseForm()
        if err != nil {
            http.Error(w, err.Error(), http.StatusInternalServerError)
            return
        }
        param := Holder{}

        if err := datastar.ReadSignals(r, &param); err != nil {
            http.Error(w, err.Error(), http.StatusInternalServerError)
            return
        }
        log.Printf("holder %+v", param)

    })
    http.ListenAndServe(":8088", router)
}

bencroker:
Thanks! This is, as we call it in the industry, a bug.

@delaneyj it appears to be caused by the `upsert` function only writing a value if the current value is `null` or `undefined`. In this case, `filesMimes` and `filesNames` are empty arrays, so no values are written. 
[...]

PR #432, opened 2024-12-29
Fix file upload signals

iansmith:
I believe this fixes the problem with file upload.  It fixed my problem, FWIW.


bencroker:
It does fix it, but what does it break? Assigning to @delaneyj to see if this is the right approach. The issue is at https://github.com/starfederation/datastar/issues/426.

delaneyj:
This will 💯 break most upsets as you are literally change the upset to an update but I can fix your specific issue

delaneyj (approved):
For posterity, I got this change wrong.  This is :100: legitimate, is was looking at the wrong area.  Good job

PR #454, opened 2025-01-05
Chore/java sdk updates

rphumulock:
- Updated the README for the java SDK to match the current implementation
- Added retryDuration to the SDK in ServerSentEventGenerator


bencroker:
Merged, thanks!

PR #458, opened 2025-01-05
Chore/java sdk submodules

rphumulock:
- Added a bit more explanation to readme
- moved sdk into submodule
- fixed a typo that could cause a bug


bencroker:
Merged, thanks!

PR #459, opened 2025-01-06
feat: Better signal errors

### LINKED ISSUE #452: Better error messages for nested signals
If you do something like this:

```html
                    <button class="btn btn-neutral"
                            data-on-click="$pokemonEntry.name = $origPokemonEntry.name"
                    >
                        Cancel
                    </button>
```
...but the nested signal `pokemonEntry.name` doesn't exist (though the signal `pokemonEntry` _does_ exist), the error message is:

```
VM380:5 Uncaught ReferenceError: $pokemonEntry is not defined
    at eval (eval at #l (engine.ts:275:18), <anonymous>:5:1)
    at eval (eval at #l (engine.ts:275:18), <anonymous>:6:3)
    at c.error (engine.ts:276:34)
    at HTMLButtonElement.l (on.ts:36:7)
```

...which isn't particularly helpful or accurate.

A better error message would be something that mentions that `name` does not exist.

khalwat:
This PR adds better error reporting for signals, and closes: https://github.com/starfederation/datastar/issues/452

Right now if you do something like this:

```html
	    <div data-signals="{foo: {bar: 'hello'}, baz: 1}">
	    <h1 data-text="$foo.bar"></h1>
	    <h1 data-text="$foo.woof"></h1>
```

You'll receive the JavaScript console error:

```
$foo is not defined
```

...which isn't very helpful to the developer. The error, of course, is that `$foo.woof` doesn't exist.

What this PR does is it checks to see if the `userExpression` still contains a `$` (which _might_ mean an unparsed signal reference) after replacing declarative signals with `ctx.signals.signal('signalname').value` and then looks to see if the `userExpression` contains any of the root properties of the `SignalsRoot` object.

That would mean that they are trying to use signals, but have got something wrong. If so, it throws an error like this that tells them what the actual error is:

Screenshot 2025-01-05 at 19 30 23

Feel free to reject the PR if it isn't written the way you'd do it... just some food for thought, and my take on it.

### DIFF
--- a/library/src/engine/engine.ts
+++ b/library/src/engine/engine.ts
@@ -4,6 +4,7 @@ import { effect } from '../vendored/preact-core'
 import { DSP, DSS, VERSION } from './consts'
 import { dsErr } from './errors'
 import { SignalsRoot } from './nestedSignals'
+import { trimDollarSignPrefix } from '../utils/text'
 import {
   type ActionPlugin,
   type ActionPlugins,
@@ -270,7 +271,7 @@ export class Engine {
       const rootProps = ctx.signals.rootProps()
       const rootPropsRe = new RegExp(`\\$(${rootProps.join('|')})(.?((\\w+).)*(\\w+)|$)`, 'gm')
       const found = userExpression.match(rootPropsRe);
-      if (found) throw dsErr(`SignalNotFound: ${found[0]}`, { path: found[0] })
+      if (found) throw dsErr(`SignalNotFound`, { path: trimDollarSignPrefix(found[0]) })
     }
     // Replace any escaped values
     for (const [k, v] of escaped) {
--- a/library/src/engine/engine.ts
+++ b/library/src/engine/engine.ts
@@ -253,6 +253,7 @@ export class Engine {
     )
 
     // Replace any signal calls
+    let origUserExpression = userExpression;
     const signalNames = new Array<string>()
     ctx.signals.walk((path) => signalNames.push(path))
     if (signalNames.length) {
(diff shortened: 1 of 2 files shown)


bencroker:
Interesting approach, although it looks like you’re attempting to catch a specific use-case. Is the following example accounted for?

```html
<div data-signals="{foo: {bar: 'hello'}, baz: 1}">
<h1 data-text="$foo.bar + $foo.woof"></h1>
```

khalwat:
> Interesting approach, although it looks like you’re attempting to catch a specific use-case. Is the following example accounted for?
> 
> ```
> <div data-signals="{foo: {bar: 'hello'}, baz: 1}">
> <h1 data-text="$foo.bar + $foo.woof"></h1>
> ```

It's not looking for a specific use-case. It's looking to see if the `userExpression` has any of the root-level properties from `SignalsRoot` (prefixed by a `$`) still in it _after_ Datastar has replaced any valid declarative signal references with `ctx.signals.signal()`. So it's generic.

I had to fix an unrelated issue that your example brought to the fore, but this works too. It results in:

Screenshot 2025-01-05 at 19 44 40

bencroker:
> It's not looking for a specific use-case.

Your original code was looking only at expressions that had not changed.

> It's looking to see if the `userExpression` has any of the root-level properties from `SignalsRoot` (prefixed by a `$`) still in it after Datastar has replaced any valid declarative signal references with `ctx.signals.signal()`.

I think you’re close, but I suspect there are still edge-cases that your code doesn’t account for. Does this return the expected error, for example?

```html
<div data-signals="{foo: {bar: {baz: {'hello'}}}">
<h1 data-text="$foo.length"></h1>
```

khalwat:
> Your original code was looking only at expressions that had not changed.

Right. I was assuming that the expression would change from what it was before declarative signals were parsed as an optimization to skip checking it. 

It was a logic error, I was just trying to not check for errors unless we needed to, it wasn't an attempt to solve a specific use-case.

Anyway I just tested your example above, it yes, it results in:

Screenshot 2025-01-05 at 20 04 54

...which is the same error that Datastar returns without my PR if you do that. It looks like that is parsed somewhere upstream, because it never reaches my code in that case.

Probably that's a separate issue that should be filed for Datastar, because it may be a bug.

bencroker:
Closing this in favour of https://github.com/starfederation/datastar/pull/466, in which we catch the runtime error, send the context to the backend, and parse the error to figure out the most helpful error message. Since we have the context, we also make suggestions for what we think the person might be trying to do. 

scre

PR #474, opened 2025-01-11
Remove unneeded snippet from Backend Examples page

### LINKED ISSUE #473: Remove extraneous html snippet from the Python "Backend Examples" page
The html snippet below isn't needed, so remove it from this page:
https://data-star.dev/examples/python

```html
<div class="container">
  <img src="foo.img" />
</div>
```

jmoppel:
On this page:
https://data-star.dev/examples/python

The html code block isn't needed to run the project.

```html
<div class="container">
  <img src="foo.img" />
</div>
```

Fixes #473


bencroker:
Thanks for the PR. We decided to remove the Python example in favour of the SDK examples found in https://github.com/starfederation/datastar/tree/develop/examples/python

PR #477, opened 2025-01-11
First draft of renaming nest/ed to namespace/d

andersmurphy:
After a lengthy discussion on discord it was suggested that we rename `nested` signals to `namespaced` signals. For a few reasons:

1. Nested signals are used for namespacing.
2. We don't support or plan to support root nodes being signals
3. Nested signals should not be used to recreate SPAs and are there for namespacing convenience (bulk add/remove) not lookup (read/watch).
4. The naming was causing confusion.

Not sure if we still need to specify that only nodes are signals now that nested signals are called namespaced signals.

I haven't renamed the implementation's yet as that's an implementation detail we can address later.

Link to discord conversation:
https://discord.com/channels/1296224603642925098/1296225503610671224/1327365393563783238

### DIFF
--- a/site/smoketests/persist_test.go
+++ b/site/smoketests/persist_test.go
@@ -20,7 +20,7 @@ func TestExamplePersist(t *testing.T) {
 				marshalled := fromLocalStorage.String()
 				c, err := gabs.ParseJSON([]byte(marshalled))
 				assert.NoError(t, err)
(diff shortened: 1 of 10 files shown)


delaneyj (approved):

bencroker:
Thanks for the PR! This sounds like a win. We’ll handle the internal change separately.

PR #483, opened 2025-01-14
Ts sdk snippets

Superpat:
Also improved the examples to include information on readSignals usage.


site/static/code_snippets/getting_started/setup.tssnippet
bencroker, line 1: Shouldn’t lines 15-24 be sufficient? If you reference the other SDK code snippets, they don’t show handling of the entire request. Something more thorough should ideally exists in the SDK’s readme file and, if appropriate in the `/examples` folder.

Superpat: You're not wrong, I fixed it.

bencroker:
Please rename the `ts` folders to `typescript` (and the snippets too, better to be consistent).

site/static/code_snippets/getting_started/setup.tssnippet
bencroker, line 1: Most of the code in this file is error handling. Could we remove that from the snippet and use show it in the usage section of the readme only?

Superpat: I think reading signals is part of getting started, but if you insist I will remove it. If I keep the readSignals example then I'd keep the error handling too.

Superpat: Though I could replace the check for `foo` with a reader.signals.foo!

bencroker: Reading signals should be included, but can we only keep the `reader.success` check, remove the duplication of reporting the error, and remove the second conditional for the existence of `foo`?

Superpat: Sure

bencroker: Thanks, that looks much better now!

bencroker:
Looks good, ready for me to merge?

Superpat:
Yes !

bencroker:
Merged, thanks!

bencroker:
Now I’m running it locally, I’m seeing inconsistencies with the other code snippets. Can you try to mirror this as closely as possible in a new PR please?
https://github.com/starfederation/datastar/blob/develop/site/static/code_snippets/getting_started/setup.phpsnippet

PR #487, opened 2025-01-15
Bring typescript snippets into line with other examples

Superpat:
I kept the readSignal usage example though.


Superpat:
The reason for deno.ts and node.ts in the src folder is that they're used to respond to the shell based test suite and I didnt want the examples to be burdened with that.

bencroker:
> I kept the readSignal usage example though.

Please remove it. The code snippet should be the equivalent of the other SDKs. Move any usage instructions into the readme file, please.

Superpat:
Are we sure `readSignal` usage should not be in all snippets instead ?

bencroker:
I’m the meantime I‘m more concerned with having all code snippets align than adding new code. Feel free to open an issue or PR if you think it would be worthwhile exploring.

bencroker:
Merged, thanks!

PR #488, opened 2025-01-15
convert TS examples (deno, node, sw) to use the new hello-world.html format

nickchomey:
This adds an example for the typescript sdk that has isomorphic routing and rendering for deno and a service worker. If you go offline, everything will be served from the SW. 
* Uses Hono for the routing etc... 
* Deno tasks watch for changes and use esbuild to rebuild the service worker as-needed. And optionally minify it. 

There's a few files that are probably extraneous - caddyfile, mise.toml. Not sure what to do about those since I'd like to have them in my own repo


nickchomey:
@Superpat care to give it a try? Im open to any feedback

nickchomey:
Changed to draft. Just tinkering with a merge-signals mechanism to toggle an offline indicator (rather than a kludgy, non-declarative, non-signals mechanism that I was using)

nickchomey:
Put in draft again - going to tinker some more, both for my own practice/learning, as well as to give people a more comprehensive TS example to get started with.

Superpat:
Great, I'll review / test when it's ready for merging !

nickchomey:
im going to keep tinkering here, but apart from whatever feedback you folks provide, I think it really is "done" for the purposes of the example

bencroker:
Thanks for the PR! I’d like @Superpat to review this first, since he submitted the Typescript SDK.

Superpat:
I think I'll have time to review this on friday.

Superpat:
@nickchomey I'd say my main concern right now is that I want the examples to remain simple / easy to read, full application level examples should probably live in their own repo's. I'll let you know how that goes when I try running it, but we may need to slim it down a bit.

nickchomey:
Very fair. I kind of just kept iterating out of curiosity.

Yet, it's actually pretty simple - 1 endpoint for the initial page, 1 for the sse content responses. They're in the shared-router file, which then gets imported by the deno and service-worker modules. Those just do some minimal stuff that's needed for each environment - that's the real value of this example. The rest was just reusing the stuff I did previously to test compressing sse responses. 

The most complex stuff was related to displaying a message when online/offline - it started simple (event listener for those events) but got more complicated when I found that that didn't work when the page was refreshed. It still doesn't work perfectly though. 

No one really needs to do anything other than have deno installed and run deno run dev (or whichever other task you want) 

I'm perfectly happy to remove or change anything if you think it would make for a simpler, better example. Perhaps removing minification from the build file and deno tasks? Likewise just making the offline indicator only use data-on-offline and data-on-online? 

I'll remove the caddyfile and mise.toml as it's ultimately not needed here, and I already scratched that itch re: compressing sse responses and documented it elsewhere. (also, someone just submitted a 2-line PR to Deno for an issue I created to add support for compressing text/event-stream responses. Hopefully it'll be merged soon).

I could also just completely replace the actual examples with one or two of the simple official examples, since all this really needs to do is show that d* can work in a service worker.

Let me know what you guys would like!

nickchomey:
One thought - I could, in theory, just prebuild the serviceworker.js and remove the build script. Likewise simplify the deno tasks to just one option.

Though I thought that maybe including stuff like that would make it easier for people to modify and experiment with it.

An alternative could be adding a Readme (should be done anyway) that links to a separate repo where they could get a buildable example. That seems convoluted to me though.

Anyway, again, I'm perfectly happy to make any changes

nickchomey:
Ive put it back in draft - suddenly it isn't working from the service worker for me... Perhaps something in one of the dependencies changed...

nickchomey:
Im just dumb - evidently I didnt properly test the most recent changes that I had made (adding minification of template literals) and it was removing `/n` which apparently messes with D* typescript sdk. 

@Superpat I know we had talked about minification of html when we were discussing compression. But given that compression is so effective and easy, and that minification of template literals really just amounts to removing line breaks and, perhaps, some comments which dont contribute much to filesize, I don't think its at all worth fixing the sdk to support doing this. Just compress js files and sse responses. 

As mentioned, Deno will hopefully compress event-steam soon. Here's the PR for it https://github.com/denoland/deno/pull/27776

And its dead-simple with caddy and other reverse proxies.

Superpat:
@nickchomey I was able to test this today and while the online deno example works fine (and is very good). If I set the network state to offline in my devtools the service worker doesnt work.

I've tried in both firefox and chromium. Do I need to do anything else then `deno -A deno.t`

nickchomey:
Ah, use `deno run dev` or one of the other tasks in deno.json. It'll build the service worker, which I assume has not been done. I suppose I could include that in the repo, pre built... 

Not sure if you have to run deno install first either. 

I should probably add a basic readme... 

Though, I was thinking last night that it really should be a simpler demo... I'll review it all later and submit a new commit

---

I've also discovered some gotchas with service workers - in particular, you cant do dynamic imports of scripts with `import()`. Not an issue for this example, but it creates some limitations/complications for making a more comprehensive SW-rendered app (unless you include all scripts in a single bundled sw). I've found a good workaround now.

I definitely won't be including anything like that here, but it really ought to be shared in some way 

Perhaps it would be best to just carry on, and when I'm ready to share an actual case study, I can link to a more comprehensive repo/guide/framework

Superpat:
I still cant get it to run, I ran `deno  run --unstable-sloppy-imports -A deno` and in the console I see 

`Service Worker registration failed: TypeError: Failed to register a ServiceWorker for scope ('￼http://localhost:8000/') with script ('￼http://localhost:8000/service-worker.js'): A bad HTTP response code (404) was received when fetching the script.`

Besides that, I think the demo should be a bit simpler yes, no need for the comparison between streamed / buffered (it's very interesting, but it would be more at home on a blog post or a page on the data-star website, maybe contribute it to the docs pr @bencroker is leading). Ideally it would be the same example as the deno / node examples. Though we could complexify those a little if you want so it's easier to showcase the offline behaviour. I imagine an input that takes a string, a button that says send and an output that is populated by sse with hello $string.

Also a readme would be crucial.

Superpat:
I will work on updating the other examples to the basic example I just detailed.

nickchomey:
Again, you're not building the sw - please look at deno.json and run whichever task you prefer. I use `deno run dev`

Good idea making it just show the same examples as non-service worker. I'll try to do that today.

And, yes, I said I'd made a readme.

Superpat:
Ah sorry, I thought I ran deno run dev but I ran deno run deno by accident.

So it's telling me that esbuild does not support OpenBSD (which is strange since I can run the go version of esbuild just fine). Would you be willing to use deno dnt to build the service worker instead ? You can look at `sdk/typescript/build.ts` for an example. I'd prefer to keep all the ts stuff buildable on openbsd if I'm going to be maintaining them. Also makes it more likely it will run on other fringe operating systems.

Superpat:
For the example, I want to have a single html file in 'sdk/typescript/examples/static' that is served by each example so that they only have to serve it and then expose an sse route that updates the view. 

Feel free to create that file if I dont get to it before you update your example, I'll use it for the others.

Superpat:
@nickchomey for your example best to follow whats discussed here: https://discord.com/channels/1296224603642925098/1296225449260879973/1333463946317009027 by @bencroker

nickchomey:
Put this into draft mode. I will work to convert this example and the existing deno.ts and node.ts examples to use the hello-world.html example that is now the standard for all sdks

nickchomey:
I broke something while trying to merge upstream. Im just going to close this and start a new PR branch...

PR #496, opened 2025-01-18
sdk/dotnet

SpiralOSS:
- corrections to doco
- fixed and touched up sample
- keep-alive sent only if HTTP/1.1


SpiralOSS:
@bencroker for your viewing pleasure

bencroker:
Merged, thanks!

PR #497, opened 2025-01-19
Support regex in expressions

andersmurphy:
**The issue:**

I was running this code to pull out a csrf cookie and add it to signals:

`document.cookie.match(/(^| )__Host-csrf=([^;]+)/)?.[2]`

Problem was it resulted in this:

```
return (()=> {
document.cookie.match(/(^| )__Host-csrf=([^;
return (]+)/)?.[2]);
})()
```

```
SyntaxError: unterminated regular expression literal
```

Because the expression regex was splitting my regex on `;`.

**The fix (more regex):**

Splitting on:

`/(?:;|\n)(?=(?:[^\/]*\/[^\/]*\/)*[^\/]*$)/`

Instead of:

`;|\n`

**Breakdown:**

- `(?:;|\n)` matches `;` or `\n` not a capture group (match existing behaviour) 
- `(?=` start positive look ahead
- `(?:[^\/]*\/[^\/]*\/)*` any number of /regex/ zero or more times 
- `[^\/]*` match any number of non `/`
- `$` match to end of string
- `)` stop positive look ahead

I appreciate this is adding more regex to the regex because we like regex. I understand the above regex, but it is still regex.

I haven't pulled out the regex into a `new RegExp` as it's only used in one place. I might be wrong but from what I remember there's no perf difference between regex literals and `new RegExp`.

Anyway happy to make any changes, or for you to rewrite this, or to reject it as insanity.

### DIFF
--- a/library/src/engine/engine.ts
+++ b/library/src/engine/engine.ts
@@ -216,7 +216,8 @@ export class Engine {
     ...argNames: string[]
   ): RuntimeExpressionFunction {
     const stmts = ctx.value
-      .split(/;|\n/)
+       // split on ; or \n but not if they are within a regex
+      .split(/(?:;|\n)(?=(?:[^\/]*\/[^\/]*\/)*[^\/]*$)/)
       .map((s) => s.trim())
       .filter((s) => s !== '')
     const lastIdx = stmts.length - 1


bencroker:
I generally don‘t like the idea of fixing a regex with an even longer regex. The current implementation is straightforward yet oversimplified – it does not account for semicolons in regular expressions, but it doesn’t account for them in strings either. Trying to account for all edge-cases using regular expressions sounds like a lost cause. Anyone got a better idea?!?

For your use-case, I’d suggest defining your regex in a variable outside of the expression, as an immediate fix:

```html
document.cookie.match(regex)?.[2]
```

bencroker:
Also not a great solution but, out of interest, have you tried encoding the semicolon in your regex?

```
document.cookie.match(/(^| )__Host-csrf=([^\x3B]+)/)?.[2]/\x3B/
```

andersmurphy:
Thanks for the feedback Ben! 🚀 

> For your use-case, I’d suggest defining your regex in a variable outside of the expression, as an immediate fix:

I've manually patched my datastar.js to test this anyway so I'm all good on my end. Although I'm trying to avoid adding another separate script, pulling this out into a function that can be used in the expression is not the worst compromise. 

> Also not a great solution but, out of interest, have you tried encoding the semicolon in your regex?

Encoding doesn't work.

Unfortunately, What I've realised though is datastar has the same issue with strings:

The expression "'hello \n world'" becomes:

```
return (()=> {
'hello;
return (world');
})()
```

While I agree with the regex sentiment. I do think regex and strings that contain `\n` or `;` in expressions have a reasonable chance of being used. This issue definetly makes the expression abstraction more leaky so could be quite a surprise to datastar users.

Got a few ideas. So looking into a cleaner fix that handles strings too. I'll post it here if I find something half decent.

However, I understand the complexity trade-off so no worries if you decide it's not worth it. I have zero expectation anything I come up with gets merged. Thanks again for the feedback.

andersmurphy:
Some more thought on this.

Behold enlightenment.

`/\/[^\/]*\/|'[^']*'|([;\n])/`

Breakdown:

`\/[^\/]*\/` We ignore everything between `/` `/`

`'[^']*'` We ignore everything between `'` `'`

`([;\n])` We want `;` or `\n` after ignoring all the things.

This blog post has changed my life:

https://www.rexegg.com/regex-best-trick.php#thetrick

The nice thing with this technique is it makes the regex trivial to understand and it handles all the edge cases. The downside is it's great for matching but doesn't play nice with javascripts slpit/replace. Because you can't just modify groups. 

```
const stmts = ctx.value
      .replace(/\/[^\/]*\/|'[^']*'|([;\n])/g, (m, g1) => (g1 == undefined) ?  m : 'H3R3CY')
      .split('H3R3CY')
      .map((s) => s.trim())
      .filter((s) => s !== '')
    const lastIdx = stmts.length - 1
    const last = stmts[lastIdx]
    if (!last.startsWith('return')) {
      stmts[lastIdx] = `return (${stmts[lastIdx]});`
    }
    let userExpression = stmts.join(';\n').trim()
```

`H3R3CY` is just something that is unlikely to collide could be `🖕JS`.

Of course we could write a regex to "just" replace the last statement with return (last statement);. Replacing all of the above code as that's what that code does (+/- some trimming that probably isn't needed). I have yet to find a readable way of doing this though. I'll meditate on it some more.

> Your mistake was not that you reached for regex... but that you didn't reach for it sooner! - Ancient Perl Proverb

bencroker:
This is very clever. I hope you like playing Whack-a-Mole! I can keep pointing out edge-cases, and we can probably reach a point where we’ve addressed _most_ of them, but I’m not sure.

Easy edge-case: **Use of strings wrapped in backticks**

```
`hello; world`
```

Difficult edge-case: **Use of escaped single quotes in strings wrapped in single quotes**

```
'\'hello; world\''
```

I could go on, but I think you’ll find that there is no regular expression that solves this entirely. At some point we’ll need to capitulate and document what is allowed in Datastar expressions and what is not.

PR #500, opened 2025-01-20
Simplify making the last statement an expression a return

andersmurphy:
From the conversation on discord `replace` can be used to simplify making the last statement in an expression a return.

Note:

- This version doesn't do any trimming.
- This does not fix #497

Ran tests and build step locally, but wasn't sure if PRs should only be the code change (rather than artefacts).


delaneyj (approved):

bencroker:
Reverted in https://github.com/starfederation/datastar/commit/91483ab6ff7087c11f71d772c22c76f7ce580f8d.

PR #504, opened 2025-01-21
Add `data-on-interval` and `__delay` modifier

bencroker:
Adds a `data-on-interval` attribute and a `__delay` modifier.

```html
<div data-on-interval__delay.1s="alert()"></div>

<div data-on-load__delay.1s="alert()"></div>
```


site/static/md/reference/attribute_plugins.md
```diff
@@ -192,6 +193,9 @@ Modifiers allow you to modify behavior when events are triggered. Some modifiers
 - `__once` \* - Only trigger the event listener once.
 - `__passive` \* - Do not call `preventDefault` on the event listener.
 - `__capture` \* - Use a capture event listener.
+- `__delay` - Delay the event listener.
+  - `.500ms` - Delay for 500 milliseconds.
+  - `.1s` - Delay for 1 second.
```
delaneyj, line 198: Do we need a withLeading so someone could initiate a call immediately instead of waiting for the interval?

delaneyj (changes requested):

site/static/md/reference/attribute_plugins.md
bencroker: I added a `.leading` option to `__delay`, which essentially nullifies the delay time and makes it so that the callable is executed immediately, _and_ at the interval determined by `__delay`. The reason the implementation is somewhat conflated is that `__delay` has a duel purpose – it delays the execution of the callable and is used as the interval duration.

delaneyj (approved):

PR #521, opened 2025-01-24
sdk/test error message improvements

Superpat:
Improved the error messages to make it clearer which errors are the fault of the user.


bencroker:
Merged, thanks!

PR #522, opened 2025-01-24
Add search to the website

Joshswooft:
Adds search functionality to the data-star website for both desktop and mobile


delaneyj (approved):

bencroker:
Thanks for this! Running locally, I see `"Search feature enabled"` but I’m not seeing any search UI. Is there anything I need to do, or just be more patient?

Joshswooft:
> Thanks for this! Running locally, I see `"Search feature enabled"` but I’m not seeing any search UI. Is there anything I need to do, or just be more patient?

Hey @bencroker , I've just added some desktop styling. I'm hoping mobile can wait until I get some more free time although I don't know when that will be.

@delaneyj The search works but already I've noticed it would benefit if for instance the title field was 'boosted' compared to the rest of the other fields? Worth playing around with but again feel like this is more of an optimization that can come in the future :)

bencroker:
Ok, thanks for your work on this thus far, and I look forward to playing with it some more.

site/shared.templ
```diff
@@ -101,6 +101,90 @@ templ headerTopLevelLinks(r *http.Request) {
...
+	<div id="header_site_search" class="hidden lg:flex lg:gap-4 relative z-10">
+		@SiteSearch(&SiteSearchSignals{
+			Search: signals.Search,
+			SearchFetching: signals.SearchFetching,
+			OpenSearchResults: signals.OpenSearchResults,
+		}, searchResults, true)
+	</div>
+}
+
+templ DrawerSiteSearch(signals *SiteSearchSignals, searchResults []SearchResult) {
```
bencroker, line 181: Is `DrawerSiteSearch` being used and, if so, could you explain what for?

site/shared.templ
```diff
@@ -101,6 +101,90 @@ templ headerTopLevelLinks(r *http.Request) {
...
+				type="search"
+				placeholder="Search..."
+				data-bind="siteSearch"
+				data-on-input__debounce.300ms={ fmt.Sprintf(
+    				"$siteSearch.length > 1 && @get('/api/search', %s)", 
+   					 fmt.Sprintf(`{"headers": {"%s": "%t"}}`, SearchFromHeaderKey, fromHeader),
+				)}
+				data-indicator="$searchFetching"
+			/>
+			@sseIndicator("searchFetching")
```
bencroker: Can you please style the indicator so this it doesn’t cause jank due to its horizontal space when visible?

screenshot-LRmqRfTQ@2x

site/shared.templ
bencroker: Ah, it appears to be be used for the mobile search. Is that correct?

bencroker: I’m guessing that the two `input` fields with `id="site_search_input"` is intentional too?

bencroker:
This is looking great! I added a keyboard shortcut to add focus and adjusted the height. Could you please address the indicator jank issue and then I think we can merge!

site/shared.templ
Joshswooft: Hey, yes the drawer site search is for the mobile input. And I've used a header value to provide a switch so it sends down the correct fragment. I'll take another look tomorrow as I'm out atm. It might be the case that the ID is not needed anymore when I refactored/added mobile.

site/shared.templ
Joshswooft: Fixed in: https://github.com/starfederation/datastar/pull/522/commits/ee297003c31024a05dae42f27b4e1ddd7f6a6fd8

site/shared.templ
Joshswooft: Just confirmed those ID's aren't needed anymore and removed them

bencroker:
Merged, thank you so much!

PR #525, opened 2025-01-25
Only support `;` statement delimiter

andersmurphy:
This means we don't need to write a JS parser in regex and keeps things simple.

I haven't updated the docs to mention this or added specific error handling. Figured that could be done in a separate PR.

This PR does fix all the tests where \n was used as a statement delimiter.


delaneyj (approved):

delaneyj (approved):

bencroker:
Merged, thanks!

PR #526, opened 2025-01-25
sdk/test eventId should be string and send datastar-request header

Superpat:
Should fix the problems the rust sdk was having because of it's stricter types


bencroker:
Merged, thanks!

PR #528, opened 2025-01-26
Revamp apply and cleanup to be based primarily on MutationObserver

delaneyj:


library/src/engine/engine.ts
```diff
@@ -14,19 +14,99 @@ import {
...
+    const ob = new MutationObserver((mutations) => {
+      for (const {
+        target,
+        type,
+        attributeName,
+        oldValue,
+        addedNodes,
+        removedNodes,
+      } of mutations) {
+        const elify = (n: Node) => n as HTMLorSVGElement
```
Regaez: Don't know if you're interested in comments on this PR, nor do I know if you would find this helpful or not, but I would leave this comment here for consideration:

Perhaps this function could be moved outside of the for loop so it's not redeclared every iteration, as it doesn't depend on any context of the loop? 

Or, perhaps more extreme, since all `elify` is doing is a Typescript type assertion, is it worth just doing the assertion inline in the 4(?) places `elify` output is used, and thus avoid introducing any runtime code at all? Might be a bit uglier and a slightly poorer DX, but I think there's an argument to be made that it's kind of a pointless function call beyond making the Typescript compiler happy...

delaneyj: valid point, fixed and it's a few bytes smaller!

library/src/engine/engine.ts
```diff
@@ -14,19 +15,96 @@ import {
...
+            }
+            break
+          }
+        }
+      }
+    })
+
+    ob.observe(document.body, {
+      attributes: true,
+      attributeOldValue: true,
```
tylerforesthauser, line 99: Could we not use `attributeFilter: ['data-*']` here instead of checking the attribute name up on Line 73?

andersmurphy: Assuming that we only care about `data` attributes then that would prevent us observing all the other attributes that we don't care about. It's one of the things advised in this post for performance: [Performance of MutationObserver to detect nodes in entire DOM](https://stackoverflow.com/questions/31659567/performance-of-mutationobserver-to-detect-nodes-in-entire-dom).

bencroker:
@delaneyj please merge in changes to `engine.ts` from the `develop` branch, so we don’t lose them.

bencroker:
A nice side effect of this is that it might help pave the way to implement https://github.com/starfederation/datastar/discussions/323 in a more elegant way.

library/src/engine/engine.ts
```diff
@@ -14,19 +15,96 @@ import {
...
+        type,
+        attributeName,
+        oldValue,
+        addedNodes,
+        removedNodes,
+      } of mutations) {
+        switch (type) {
+          case 'childList':
+            {
+              if (removedNodes.length) {
```
Regaez: Are these wrapping if statements redundant considering that, if `removedNodes` or `addedNodes` have zero length, the for loops within wouldn't iterate over anything?

library/src/engine/engine.ts
```diff
@@ -83,117 +159,120 @@ export class Engine {
...
+    // Extract the key and value from the dataset
+    let [key, ...rawModifiers] = rawKey.slice(plugin.name.length).split(/\_\_+/)
+
+    const hasKey = key.length > 0
+    if (hasKey) {
+      // Keys starting with a dash are not converted to camel case in the dataset
+      const keySlice1 = key.slice(1)
+      key = key.startsWith('-') ? keySlice1 : key[0].toLowerCase() + keySlice1
+    }
+    const value = `${el.dataset[rawKey]}` || ''
```
Regaez: If `el.dataset[rawKey]` is undefined, the string interpolation would result in a value of `"undefined"` and not an empty string. Is that intended?

Perhaps this is desired instead:
```suggestion
    const value = el.dataset[rawKey] || ''
```

library/src/engine/engine.ts
delaneyj: True, it's vestigial from another change that got pulled out

library/src/engine/engine.ts
delaneyj: it takes keys, not regex unforturnately

tylerforesthauser: Dang

bencroker:
Please merge `develop` and update the bundler accordingly.
https://github.com/starfederation/datastar/blob/develop/site/routes_bundler.qtpl

PR #530, opened 2025-01-26
Fix signal value out of sync when input populated by chrome's back/forward cache

td0m:
<details>
<summary>Code</summary>

in `/foo.html`:

```
<!DOCTYPE html>
<html>
  <head>
    <title>Sandbox</title>
    <meta charset="UTF-8" />
    <script
      type="module"
      src="https://cdn.jsdelivr.net/gh/starfederation/datastar@v1.0.0-beta.1/bundles/datastar.js"
    ></script>
  </head>

  <body>
    <input data-bind-title />
    <div data-text="$title"></div>
    <button data-on-click="location.href = '/'">submit</button>
  </body>
</html>
```
</details>

Visit `/foo.html`, type "bar" into the textbox, and submit. You'll get redirected to `/`. Now press the "back" button to go to the previous page (`/foo.html`).

## Expected

I would expect both to be in sync. either:
 - a blank input, or
 - an input prefilled with the back/forward cache, and the div below to contain the same value (personally I think this is the better option)

## Actual

Out of sync in Chrome (input shows `bar` but signal is empty). Not an issue in FF since they don't carry back/forward cache in this case.

## Solution

This is caused by the [back/forward cache](https://web.dev/articles/bfcache).

I've fixed the issue by updating the value of the signal if it was already populated by bfcache. Not sure if this could cause issues/conflicts with `data-signals-title="default value"`. Lmk what you think.


bencroker:
This looks great, thanks for the PR! Please review, @delaneyj.

library/src/plugins/official/dom/attributes/bind.ts
delaneyj, line 1: The looks like a totally valid change.  Pleas add full commented documentation of the reasoning and the links to the bfcache so it doesn't get lost in the future!

delaneyj (changes requested):

bencroker:
Please add a release note at https://github.com/starfederation/datastar/blob/develop/CHANGELOG.md

bencroker:
Great work, thanks!

PR #533, opened 2025-01-27
Fix recurring back through a ancestors to find an id to hash

andersmurphy:
When we recur through ancestors of a node to generate an id we were skipping back two parent nodes instead of one on the most common branch in the recursion.

This would mean html structure like this would not produce unique ids:

```html
...
<div id="3">
    <div>
        <div>
            <div id="datastar1pyvig">hello</div>
            <div id="datastargves4o">hello</div>
        </div>
    </div>
</div>
<div id="4">
    <div>
        <div>
            <div id="datastar1pyvig">hello</div>
            <div id="datastargves4o">hello</div>
        </div>
    </div>
</div>
...
```

Where as html structure like this would (even layer of nesting):

```html
...
<div id="1">
    <div>
        <div>
            <div>
                <div id="datastar1d">hello</div>
                <div id="datastarsx">hello</div>
            </div>
        </div>
    </div>
</div>
<div id="2">
    <div>
        <div>
            <div>
                <div id="datastar1t">hello</div>
                <div id="datastarsr">hello</div>
            </div>
        </div>
    </div>
</div>
...
```

Ideally we want to find any unique user generated id when recurring back through ancestors to avoid duplicate ids. Which looks like the original intention of this code (as that way users don't have to manually give ids to inner elements like buttons).

Would be good to have a look over this. Just to make sure I'm not missing any context. As far as I can tell there was no reason for the extra parent call.


delaneyj (approved):

bencroker:
Looks good! Can you please add a release note to https://github.com/starfederation/datastar/blob/develop/CHANGELOG.md ?

bencroker:
Merged, thanks!

PR #534, opened 2025-01-27
Fix custom validity plugin being too restrictive

td0m:
# Expected Behavior

Allow setting `data-custom-validity` for all valid elements, like `textarea`, not just inputs.

# Actual Behavior
I encountered the `CustomValidityInvalidElement` error while trying to set custom validity on a `textarea`. Turns out datastar only allows doing this on `input`s. 

# Solution
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity), `setCustomValidity` can be used on any `HTMLObjectElement`. I've tweaked the code to reflect this.


bencroker:
Merged, thanks!

bencroker:
FYI this introduced a bug, which I’ve fixed in https://github.com/starfederation/datastar/pull/544.

PR #541, opened 2025-01-28
Add .NET MVC Active Search Example

RickTheHat:
1. Introduce an example implementation demonstrating active search functionality in a .NET MVC application with DataStar.
2. Include relevant controllers, views, and models to handle search input and return filtered results.
3. Project is runnable with the command dotnet watch and see the demo .


bencroker:
> Help Wanted!
> We're seeking contributors who:

What exactly is the goal of this PR? It feels like it should be in a repo that you control, otherwise every contribution will have to go through the Datastar maintainers.

RickTheHat:
@bencroker I thought the team was looking for more "working demos" to go with the sdk but I *think* I hear what you're say'in that this example will need mainteance (which we don't want extra of) so I can certainly move this to a personal domain so in that case, can we put a link to where I end up hosting it?

bencroker:
Sure, we can link to it from the SDK reference page. The SDKs can provide sample usage examples, but fully featured examples should live in their own third-party repos, such as https://github.com/delaneyj/1a5s-datastar and https://github.com/khalwat/craft-datastar-pokemon-demo.

PR #546, opened 2025-01-28
Zig SDK framework integration + hello-world examples

ghost:
Adds framework support for http.zig and tokamak and hello-world examples for each framework.


examples/zig/httpz/hello-world/src/hello-world.html
bencroker, line 1: This should be built by running `task build` or `make dev`.

examples/zig/tokamak/hello-world/hello-world.html
bencroker, line 1: This should be built by running `task build` or `make dev`.

bencroker (approved):

bencroker:
Merged, thanks!

PR #548, opened 2025-01-28
Incorrect hello-world path for zig tokamak example

ghost:
Incorrect path for the generated hello-world example, embarrassed I didn't catch this!


bencroker:
Merged, thanks!

PR #550, opened 2025-01-29
sdk/test improving formatting and handle defaults better

Superpat:
The test suite was not handling defaults in the right way, they're supposed to be elided.  Add new tests to check that the response is correct in cases where all defaults are provided and cases where they are not.

Also formats the input's with jq(1) since it got too complicated to edit.


Superpat:
Ahh damnit there are merge artefacts in there, give me a second

Superpat:
@bencroker should be all good

bencroker:
Merged, thanks!

PR #551, opened 2025-01-29
Sdk ts web improvements

### LINKED ISSUE #515: Web Stream Improvements in TS SDK
I was taking a look through the web [serverSentEventGenerator](https://github.com/starfederation/datastar/blob/develop/sdk/typescript/src/web/serverSentEventGenerator.ts) and noticed some potential improvements which could broaden the appeal of the SDK.

1. The `stream` static function does not accept a `Response` object, so it's not possible for the user to modify any additional parameters on SSE responses. This could be resolved by accepting an optional `Response` object, and setting the necessary SSE headers.
2. There doesn't seem to be any error handling around the user provided stream function.
3. What happens if a stream is aborted?

Some inspiration could be taken from:
https://github.com/honojs/hono/blob/ecf5a0b2aeaf29c26ffe1eaf41520f867510cb17/src/utils/stream.ts
https://github.com/honojs/hono/blob/ecf5a0b2aeaf29c26ffe1eaf41520f867510cb17/src/helper/streaming/sse.ts

Superpat:
An attempt to fix #515 

I'm open to suggestions. I havent added abort handling yet to the web generator, because I wanted feedback. 

Up till now my deno servers havent been giving errors when client's abort, but I havent had much cleanup to do either. I know user's can use `req.onabort` to react to aborted clients, but should I be calling that in the sdk and asking users for an optional onAbort callback or should I just let users handle that. 

This pr also adds in default eliding.

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -41,9 +41,9 @@ Follow the links for more complete (and executable) examples
 
 ## Frameworks / Alternate runtimes
 
-If you can't simply use the node / web versions, then you can extend the abstract
-class in `./src/abstractServerSentEventGenerator.ts`. You will need to provide
-implementations of the `constructor`, `readSignals`, `stream` and `send`
+If you can't simply use the node / web versions, then you can extend the
+abstract class in `./src/abstractServerSentEventGenerator.ts`. You will need to
+provide implementations of the `constructor`, `readSignals`, `stream` and `send`
 methods.
 
 ## Testing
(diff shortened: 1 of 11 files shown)


Superpat:
You can merge this, I will do a separate pr for node improvements and for bringing it in line with the sdk standard and move my stricter functions into a sugared version. At which point we could do a another release.

Superpat:
Should partly fix #515  node fixes will be handled in another pr

bencroker:
Merged, thanks!

PR #552, opened 2025-01-30
deno, node and deno-sw hello world examples

nickchomey:
Here's the deno, node and deno-sw typescript sdk examples. 

I had to make local copies of the static assets for the deno-sw example, as I was having difficulty with caching and serving external resources. Likewise, the html doc is served from an ES Module that contains a template literal of the doc, as I was having a lot of issues with serving a static file as the root route (`"/"`). 

If you want/know how to fix any of that, feel free to suggest the required changes. Otherwise, I think its fine as-is for the sake of a demo.


nickchomey:
@Superpat this is ready for your review whenever you've got time

Superpat:
I'm not sure I'll have time to review this until after Fosdem, but thank you !

Superpat:
@nickchomey like I said in an earlier pr, I cant use the js version of esbuild  on OpenBSD, so I cant review the sw bits. Can you switch the sw building code to use something like deno dnt instead ? Check the npm package build code, it's using deno dnt.

examples/typescript/deno-sw/README.md
```diff
@@ -0,0 +1,9 @@
+This is a sample app that can serve the same responses from both the Deno backend and the Service Worker while offline. 
+
+To test it, start the server with `deno run -A deno.ts` and load the site at `http://localhost:8000`.
+
+The service worker will install and static assets will be cached. You can then either turn on offline mode in your browser's dev tools Network tab, or turn off the Deno server (there will be a slight delay in the response as it tries to fetch from the network first). Reload the page while offline and it will load from the service worker! And the SSE response when you click Start will be rendered from the SW as well.
```
Superpat: The service worker will be built

examples/typescript/deno-sw/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+If you want to expand upon it, it would be best to modify the `shared-router.ts` file, which is what both versions ultimately use (other than for loading static assets, which require different mechanisms in each environment).
+
+There's many other things that can be done from a service worker - including different caching and network strategies. [Google Workbox](https://developer.chrome.com/docs/workbox) is a good resource - both as an easy-to-use library as well as just reference material for all things PWA.
```
Superpat: I'd prefer not linking with google websites. What do you think @bencroker am I being anal ?

examples/typescript/deno-sw/build.ts
Superpat, line 1: Like I said on the pr, can you move this to deno dnt instead please ?

examples/typescript/deno-sw/deno.ts
```diff
@@ -0,0 +1,25 @@
+import { Hono } from "jsr:@hono/hono/tiny";
+import { serveStatic } from "jsr:@hono/hono/deno";
+import { createRouter } from "./shared-router.ts";
+
+// Rebuild service worker first
+const buildProcess = new Deno.Command(Deno.execPath(), {
+  args: ["run", "-A", "--unstable-sloppy-imports", "build.js"],
+});
+await buildProcess.output();
```
Superpat: You should probably check the output and block further execution if it did not work. This is what confused me when reviewing your last pr.

examples/typescript/deno-sw/static/rocket.png
Superpat, line 1: @bencroker  should he be reusing the rocket image from the go websites static files or is it fine to reupload it here ?

Superpat (changes requested):
Globally it looks good, I'll want to run it first, but I need you to do a few changes as delineated in the comments.

nickchomey:
Sure I'll figure out deno dnt

examples/typescript/deno-sw/static/rocket.png
nickchomey: I'll try again to get the external asset caching working. I have a suspicion as to what the problem was. Though the tailwind, in particular, was partly not working because it does a couple redirects from the url that is being used.

examples/typescript/deno-sw/README.md
nickchomey: Woild it be better to link to the Github repo? It's not strictly necessary to provide this info at all, but I figured id at least point people in a helpful direction.

bencroker: I guess it depends on how important of a resource it is. Let’s remove it for now, we don’t need to recommend other libraries.

examples/typescript/deno-sw/static/rocket.png
bencroker: If we can avoid duplicating assets, that would be preferable.

nickchomey:
@Superpat 
I got the static assets to load remotely now, tidied things up, and have provided a pre-built service-worker.js file as I wasted an hour trying to get DNT to work - its not worth the effort for this. 

You can just run `deno run start` and it should just work. If you/anyone wants to modify the files and doesnt have esbuild available on their system, then they'll have to figure out a build system that works for them. 

The only remaining "gotcha" is a couple second delay if your browser is still set to online, but the network request fails (eg deno is stopped) - it waits for that request to fail before falling back to service worker rendering. If you set the browser to offline, it goes there immediately. I could add some sort of background polling and a signal to pre-detect if the network/deno server is available, but that really isn't worthwhile for this proof-of-concept. 

ps. note that I've removed the sdk/typescript/examples directory

examples/typescript/deno-sw/static/rocket.png
nickchomey: I got this working now - everything is served remotely and cached for use when offline.

examples/typescript/deno-sw/deno.ts
nickchomey: Ive just removed this altogether. Theres deno tasks for start, build and dev (build + start). Its using Esbuild still because, as mentioned in a top-level comment here, i cant get dnt to work. But ive included a pre-built sw.js so you dont need to build - just `deno run start`

Superpat:
I will try to finish reviewing this before the weekend is over. Still recovering from my ssd meltdown last week.

Superpat:
@nickchomey can you move the examples over to https://github.com/starfederation/datastar/tree/develop/examples

delaneyj:
It's been 84 years

bencroker:
It’s been a light year...

Issue #516, opened 2025-01-23
How-to Guides


bencroker:
Dear community, we need **HELP** transforming the [Examples](https://data-star.dev/examples/click_to_edit) section of the Datastar website into [How Tos](https://data-star.dev/how_tos).

![Image](https://github.com/user-attachments/assets/b4beee0a-7590-4318-976b-4d418ab0f9da)

While the existing examples cover a lot of use-cases, they are unstructured, incomplete, poorly explained, and they would provide much more value as goal-oriented “recipes” that guide the reader towards a result.

Examples converted to How-to guides might read something like:

_Click to Edit → Inline editing on click_
_Bulk Update → Bulk updating an item list_
_Click to Load → Load more results on click_
_Delete Row → Deleting a row in a table_
_..._
_Polling → Polling the backend at regular intervals_
_Bind Keys → Binding keydown events to specific keys_
_Classes → Toggling classes based on expressions_
_CSRF → Sending a CSRF token via a request header_

Inspiration for this is taken from [Diátaxis](https://diataxis.fr/), specifically the section on [How-to guides](https://diataxis.fr/how-to-guides/) (please read _before_ deciding to contribute).

If you would like to contribute, first check out the existing [how tos](https://data-star.dev/how_tos). You can propose a how-to guide that you’d like to write in a comment in **this issue**. Please propose _one_ how-to guide at a time, and _only_ propose a guide that you are willing to contribute a complete PR for. 

To discuss ideas for how-to guides, please use the [#documentation](https://discord.com/channels/1296224603642925098/1332026342962630769) channel in the Datastar [Discord server](https://discord.gg/bnRNgZjgPh).

You should ideally be familiar with Datastar v1.0.0-beta.x and have already used it to build something. You should be comfortable writing markdown and willing to write some Go code for examples that require requests to the backend.

Proposals must follow this format:

---

TITLE: _The title of the how-to guide you are proposing_
SUMMARY: _A brief summary of what it will cover_
CODE SAMPLE: _A code sample that it might use_

---

If the proposal adheres to these [key principles](https://diataxis.fr/how-to-guides/#key-principles) and generally aligns with the goal of this initiative, a file will be added to the PR and assigned to you. You can then fork the PR and work on your how-to guide in your own forked branch. When you feel it is ready for review, submit a pull request to [PR #517](https://github.com/starfederation/datastar/pull/517).

Please note that in the spirit of FOSS (free and open-source software), we reserve the freedom to decline suggestions and pull requests.

[...]

PR #553, opened 2025-01-30
Creates an example of Optimstic Update using Datastar

schreiaj:
This is a common enough question that it comes up. It is a bad UX pattern but the reality is that it seems to be fairly used pattern. Showing that you _can_ do it while pointing to better ways to handle it will hopefully reduce the claims of needing client side templating. 

Kapture 2025-01-30 at 01 14 54

### DIFF
--- a/site/static/md/examples/optimistic_update.md
+++ b/site/static/md/examples/optimistic_update.md
@@ -1,0 +1,51 @@
+## Optimistic Update
+
+While optimistic updates are not ideal, they are achievable with DataStar. The big drawback with this style of UX is 
+that the users will see the update happening before it actually happens. Care must be taken to indicate to the user that
+this update is not fully persisted and may still fail.
+
+See the [indicator example](/examples/indicator) for a more correct way to indicate the state of a fetch request.
+
+## Demo
+
+
+<div>
+  <div data-signals="{fetching: false, name:''}">
+    <div  class="text-primary">
+        <input id="name" data-attr-disabled="$fetching" type="text" data-bind="name">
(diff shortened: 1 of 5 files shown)


site/static/md/examples/optimistic_update.md
```diff
@@ -0,0 +1,51 @@
+## Optimistic Update
+
+While optimistic updates are not ideal, they are achievable with DataStar. The big drawback with this style of UX is 
+that the users will see the update happening before it actually happens. Care must be taken to indicate to the user that
+this update is not fully persisted and may still fail.
```
andersmurphy, line 5: I don't think this is opinionated enough. Maybe, something along the lines of:

9ig8aw

Datastar is so fast at updating the dom that you shouldn't need to do client side optimistic updates (even with high network latency). In the most extreme latency situations adding a transition or animation can be enough to keep the UI feeling snappy.

The other reason you shouldn't do optimistic updates is that they can mislead a user into thinking an action has been completed when it hasn't. Imagine thinking you have successfully booked a plane ticket, only to find out later it was an optimistic update and your booking did not go through. 

If you still want to implement the optimistic update pattern in Datastar despite these shortcomings. Here's how you can do it.

bencroker:
Thanks for the PR. As mentioned in Discord, we are phasing out the examples page and replacing it with [how-to guides](https://github.com/starfederation/datastar/issues/516). I think there is huge value in a how-to guide that discusses using loading indicators and CSS/view transitions versus optimistic UI (and showing both), and I hope that you’ll consider putting in some extra effort and rolling this into one.

PR #555, opened 2025-01-30
Bundler aliasing

### LINKED ISSUE #554: Bundler aliasing

delaneyj:
Fixes #554


bencroker:
It looks like the alias is _not_ being applied to the core plugins.

screenshot-P4CzTkwZ 2@2x

bencroker:
Based on our discussion, I changed how this works to modify the key rather than the plugin names. This way, error messages should continue to work. Added an example that uses an aliased bundle, too.

delaneyj (commented):
LTGM :shipit: 
this is evidence

PR #556, opened 2025-01-30
Kotlin SDK

MikeDepies:
This is the sdk I've been using for my Kotlin projects. 

- The sdk is (slightly) coupled with the KTOR framework, but can be separated into two modules for core and adapters to support different solutions (spring, 4khttp, ...).
- Work still remains for integrating with the test suite. 
- Wasn't sure about all of the files needed to be updated for docs, so more direction & work is required there.


bencroker:
This is looking great, @MikeDepies! Please add a Hello World example, see the [Go implementation](https://github.com/starfederation/datastar/tree/develop/examples/go/hello-world) for reference.

I’ll switch this back to draft until you’re a bit further along.

bencroker:
Was this intentionally closed, @MikeDepies?

PR #557, opened 2025-01-30
Ignore removal of signals attribute

bencroker:
This fixes signal values being set to empty, due to the attribute being removed immediately after being applied, being picked up by MutationObserver and then being reapplied with an empty value.

The issue can be observed in http://localhost:8080/examples/ignore_attributes

Since this requires special treatment, I’ve also removed the `removeOnLoad` option for plugins.


bencroker:
Closing, as this is being addressed in https://github.com/starfederation/datastar/pull/555.

PR #558, opened 2025-01-30
Rust SDK framework integration + hello-world examples

ghost:
This makes it easier to add more framework integrations + Rocket integration + hello-world examples


bencroker:
Looks great! Can you confirm that the `retry` is only sent when not set to the default, in all situations?

ghost:
Yes, retry is only written to the stream in one place in the code
```rust
let millis = self.retry.as_millis();
if millis != consts::DEFAULT_SSE_RETRY_DURATION as u128 {
        write!(f, "retry: {}\n", millis)?;
}
 ```

bencroker:
Perfect, merged, thanks!

PR #560, opened 2025-01-31
Sdk/dotnet updates

SpiralOSS:
- no longer starting streaming response if only accessing signals
- documentation updates
- signals are request-based; missing trailing newline added; refactoring
- removed all non-standard example and library code for dotnet


bencroker:
Merged, thanks!

PR #561, opened 2025-01-31
Use `retry` in SSE output examples instead of `retryDuration`

ismasan:
The SDKs expect `retryDuration` as an option, but it maps to `retry` in the SSE output.

CleanShot 2025-01-31 at 18 06 56@2x


delaneyj (approved):
Your curse is you actually read the docs! Thanks!

 :shipit:

bencroker:
Merged, thanks!

PR #570, opened 2025-02-02
Add Accept: text/event-stream to bash tests

ismasan:
Datastar responses are also valid SSE responses, and the D* client already sends `Accept: text/event-stream` to the server, so tests should replicate that.

Some servers might rely on this for content negotiation.

### DIFF
--- a/sdk/test/test-get.sh
+++ b/sdk/test/test-get.sh
@@ -7,7 +7,7 @@ input=$(cat "$1/input.json")
 
 [ -f "$1/testOutput.txt" ] && rm "$1/testOutput.txt"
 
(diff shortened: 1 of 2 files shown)


bencroker:
Please review, @Superpat.

ismasan:
Any news on this?

If not this `Accept: text/event-stream` header, could we fix the current `datastar-request` header in the test scripts?

The current use `-H "datastar-request"` just omits the header in my version of cURL (Mac).
It works if I add the explicit value, ie `-H "datastar-request: true"`

As it stands now, the test scripts don't send _any_ header to indicate these are D* requests.

Superpat:
Sorry @ismasan I've been travelling for a conference. I'll check it out later tonight.

Superpat:
@bencroker we can close this now that #607 has been merged. Thank you @ismasan for the pr.

Issue #567, opened 2025-02-01
Website: Overflow issues on mobile


axelknock:
On small screens a few of the examples will overflow the width of the screen.

Mostly examples with tables and buttons in flexboxes.

<img width="566" alt="Image" src="https://github.com/user-attachments/assets/52d37269-1edd-4fcc-a6fe-3b24db411da9" />

Solution is overflow-scroll on tables and default flex-col with md:flex-row on button containers.

bencroker:
Fixed by https://github.com/starfederation/datastar/pull/575, thanks!

PR #575, opened 2025-02-02
Fix overflow issues and add menus (#567)

axelknock:
Adds overflow-scroll to examples to avoid overflow issues on small screen sizes. Also adds default flex column with flex row for examples where buttons overflow.


bencroker:
Thanks for the PR! Is there any reason not to add `overflow-scroll` to the `article` element that wraps all examples?

axelknock:
The content that causes the overflow is usually the better target since you'll be able to scroll past content that should fit on the screen.



I tried targeting the table/tbody tag since that's where most of the overflow issues occurred, but that doesn't fix it since they all have 100% width, it's their parent that has to have the style.

bencroker:
Good point, thanks for clarifying, @axelknock.

bencroker:
Merged, thanks!

PR #578, opened 2025-02-03
Zig SDK elide retry if default

ghost:
Fixes Zig SDK always including retry duration even if it's the default duration.


bencroker:
Merged, thanks!

PR #579, opened 2025-02-03
Make SDK tests use bash exit instead of return

ismasan:
Make test bash scripts use `exit` instead of `return`. AFAIK `return` is only meant within functions.

Also make them `exit 0` on success test case (no file differences) and `exit 1` on failure (SDK output differs from expected value).

This works for me on a Mac.


bencroker:
Merged, thanks!

PR #585, opened 2025-02-03
Add dispatched `retrying` event

### LINKED ISSUE #583: Trigger SSE events for `onerror`, and other SSE lifecycle callbacks
## What

It'd be great to be able to hook into SSE lifecycle events. In particular `onerror`. This would allow the UI to react to, for example, temporary server outages, while the client is retrying the connection.

```html
<p 
  data-signals-connection="waiting"
  data-on-datastar-sse="evt.detail.type='connectionError' ? $connection = 'retrying...' : undefined"
  data-text="$connection">
</p>
```

Or perhaps suffix the event names with the event type instead of a generic `datastar-sse`, which I think could be more consistent with how custom events are handled?

```html
<p 
  data-signals-connection="waiting"
  data-on-datastar-sse-connectionerror=" $connection = 'retrying...'"
  data-on-datastar-sse-connectionopen=" $connection = 'connected'"
  data-text="$connection">
</p>
```

## Why

So that the UI provides affordances to give the user feedback on SSE connection status, especially when streaming (chats, games, live dashboards).

## How

By triggering these events from the relevant SSE callbacks. Example [onerror](https://github.com/starfederation/datastar/blob/97c0760a48db484fd97f585a33e9b7a920bab864/library/src/plugins/official/backend/actions/sse.ts#L154)

```typescript
onerror: (error) => {
  if (isWrongContent(error)) {
    // don't retry if the content-type is wrong
    throw runtimeErr('InvalidContentType', ctx, { url })
  }
  // let it retry. Emit event.
  if (error) {
    console.error(error.message)
    dispatchSSE(ERROR, { type: 'connectionerror', message: error.message })
   }
}
```

... Or with specific error types, however this is handled.

```typescript
dispatchSSE(CONNECTION_ERROR, error.message)
```

Ideally, similar events could be emitted by other lifecycle callbacks, such as `onopen`.

bencroker:
Adds a `retrying` event that is dispatched when the SSE plugin is retrying to connect.

Fixes #583.

### DIFF
--- a/library/src/plugins/official/backend/actions/sse.ts
+++ b/library/src/plugins/official/backend/actions/sse.ts
@@ -14,6 +14,7 @@ import {
   type DatastarSSEEvent,
   ERROR,
   FINISHED,
+  RETRYING,
   STARTED,
 } from '../shared'
 
@@ -152,6 +153,7 @@ export const sse = async (
         // do nothing and it will retry
         if (error) {
           console.error(error.message)
+          dispatchSSE(RETRYING, { message: error.message })
         }
       },
     }
--- a/library/src/plugins/official/backend/shared.ts
+++ b/library/src/plugins/official/backend/shared.ts
@@ -6,6 +6,7 @@ export const SWAPPING_CLASS = `${DATASTAR}-swapping`
 export const STARTED = 'started'
 export const FINISHED = 'finished'
 export const ERROR = 'error'
+export const RETRYING = 'retrying'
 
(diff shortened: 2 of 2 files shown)


ismasan:
Looks good, but do these errors always mean “retrying”? For example when max retries is set to 0.

delaneyj:
retrying means retrying, i'm good with this change

delaneyj (approved):

PR #592, opened 2025-02-04
data-on-signals-change-foo (fixes #587)

### LINKED ISSUE #587: Add optional key to `data-on-signals-change`
Add the ability to listen for when a specific signal changes (following on from https://github.com/starfederation/datastar/discussions/323).

```html
<div data-on-signals-change-foo="alert('foo changed')"></div>
```

It would be even better if `data-on-signals-change-foo` was triggered whenever `foo` or `foo.bar` changes, allowing you to listen on namespaced signal changes. 

To listen for another signal change, just add another attribute.

```html
<div data-on-signals-change-foo="alert('foo changed')"
        data-on-signals-change-baz="alert('baz changed')"
></div>
```

delaneyj:


bencroker:
It looks like modifiers no longer work with `on-signals-change`, for example: 

```html

<div data-on-signals-change__remote="...">
```

This appears to have been broken in 1.0.0-beta.3, as seen in the example at https://data-star.dev/examples/signals_changed

bencroker:
Added the ability to listen to a signal path.

Removed `__remote` from the example, as it anyway was not working nor documented.

PR #594, opened 2025-02-04
Rust and Zig SDK snippets

ghost:
Adds all snippets except for redirect 3 for the Rust and Zig sdks.


bencroker:
Merged, thanks!

PR #595, opened 2025-02-04
Update setup.tssnippet: I think it should be just json.

rofrol:


bencroker:
Thanks for the PR! I’ve addressed this in https://github.com/starfederation/datastar/pull/597, since there are more such instances that need to be fixed.

PR #600, opened 2025-02-04
Ruby SDK

ismasan:
Implement the [Datastart SSE procotocol](https://data-star.dev/reference/sse_events) in Ruby. It can be used in any Rack handler, and Rails controllers.

## Installation

For now, point your `Gemfile` to the source

```bash
gem 'datastar', git: 'https://github.com/starfederation/datastar', glob: 'sdk/ruby/*.gemspec'
```

### Initialize the Datastar dispatcher

Rails, Sinatra and other frameworks already have `request` and `response` objects in their controllers:

```ruby
datastar = Datastar.new(request:, response:, view_context: self)
```

In other Rack endpoints, you can instantiate from a Rack `env` hash with:

```ruby
datastar = Datastar.from_rack_env(env)
```

### Sending updates to the browser

There are two ways to use this gem in HTTP handlers:

* One-off responses, where you want to send a single update down to the browser.
* Streaming responses, where you want to send multiple updates down to the browser.

#### One-off update:

```ruby
datastar.merge_fragments(%(<h1 id="title">Hello, World!</h1>))
```
In this mode, the response is closed after the fragment is sent.

#### Streaming updates

```ruby
datastar.stream do |sse|
  sse.merge_fragments(%(<h1 id="title">Hello, World!</h1>))
  # Streaming multiple updates
  100.times do |i|
    sleep 1
    sse.merge_fragments(%(<h1 id="title">Hello, World #{i}!</h1>))
  end
end
```
In this mode, the response is kept open until `stream` blocks have finished.

#### Concurrent streaming blocks

Multiple `stream` blocks will be launched in threads/fibers, and will run concurrently.
Their updates are linearized and sent to the browser as they are produced.

```ruby
# Stream to the browser from two concurrent threads
datastar.stream do |sse|
  100.times do |i|
    sleep 1
    sse.merge_fragments(%(<h1 id="slow">#{i}!</h1>))
  end
end

datastar.stream do |sse|
  1000.times do |i|
    sleep 0.1
    sse.merge_fragments(%(<h1 id="fast">#{i}!</h1>))
  end
end
```

### Datastar methods

All these methods are available in both the one-off and the streaming modes.

#### `merge_fragments`
See https://data-star.dev/reference/sse_events#datastar-merge-fragments

```ruby
sse.merge_fragments(%(<div id="foo">\n<span>hello</span>\n</div>))

# or a Phlex view object
sse.merge_fragments(UserComponet.new)

# Or pass options
sse.merge_fragments(
  %(<div id="foo">\n<span>hello</span>\n</div>),
  merge_mode: 'append'
)
```

#### `remove_fragments`
 See https://data-star.dev/reference/sse_events#datastar-remove-fragments

```ruby
sse.remove_fragments('#users')
```

#### `merge_signals`
 See https://data-star.dev/reference/sse_events#datastar-merge-signals

```ruby
sse.merge_signals(count: 4, user: { name: 'John' })
```

#### `remove_signals`
 See https://data-star.dev/reference/sse_events#datastar-remove-signals

```ruby
sse.remove_signals(['user.name', 'user.email'])
```

#### `execute_script`
See https://data-star.dev/reference/sse_events#datastar-execute-script

```ruby
sse.execute_scriprt(%(alert('Hello World!'))
 ```

#### `signals`
See https://data-star.dev/guide/getting_started#data-signals

Returns signals sent by the browser.

```ruby
sse.signals # => { user: { name: 'John' } }
 ```

#### `on_connect`
Register server-side code to run when the connection is first handled.

```ruby
datastar.on_connect do
  puts 'A user has connected'
end
```

#### `on_client_disconnect`
Register server-side code to run when the connection is closed by the client

```ruby
datastar.on_client_connect do
  puts 'A user has disconnected connected'
end
```

#### `on_server_disconnect`
Register server-side code to run when the connection is closed by the server.
Ie when the served is done streaming without errors.

```ruby
datastar.on_server_connect do
  puts 'Server is done streaming'
end
```

#### `on_error`
Ruby code to handle any exceptions raised by streaming blocks.

```ruby
datastar.on_error do |exception|
  Sentry.notify(exception)
end
```
Note that this callback can be registered globally, too.

### Global configuration

```ruby
Datastar.configure do |config|
  config.on_error do |exception|
    Sentry.notify(exception)
  end
end
```

### Tests

```ruby
bundle exec rspec
```

#### Running Datastar's SDK test suite

Install dependencies.
```bash
bundle install
```

From this library's root, run the bundled-in test Rack app:

```bash
bundle puma examples/test.ru
```

Now run the test bash scripts in the `test` directory in this repo.

```bash
./test-all.sh http://localhost:9292
```


bencroker:
Thanks for the PR, looks like a great start!

Can you please follow the remaining tasks as outlined in the [Contribution Guidelines](https://github.com/starfederation/datastar/blob/main/CONTRIBUTING.md#sdks)?

I’ll put this PR into draft mode until it’s ready for review again.

ismasan:
I pushed changes to code-get constants, and a few code snippets.

I don't think existing SDKs have working examples for all code snippets. Is that a requirement for new SDKs going forward?

ismasan:
@bencroker I think I've ticked most of the contribution points. I haven't implemented all the example snippets but did add some examples that touch on streaming, fragments and signals.

I've also made the version `1.0.0.beta.1`, which in Ruby can be installed with the `--pre` flag as per convention.
This should make clear that this library needs some usage in production.

BUILDING.md
```diff
@@ -28,6 +28,13 @@ Finally, `task` will build the library and serve the reference documentation web
 
 The website should be available on http://localhost:8080 with hot-reloading enabled.
 
+The following example runs the `build` task via Docker.
+This task will rebuild SDK constants.
+
+```
+make task "build -f"
```
bencroker: This should probably go in https://github.com/starfederation/datastar/blob/develop/DOCKER.md.

build/run.go
```diff
@@ -133,6 +133,7 @@ func writeOutConsts(version string) error {
 		"sdk/go/consts.go":                                                                 goConsts,
 		"sdk/dotnet/src/Consts.fs":                                                         dotnetConsts,
 		"sdk/php/src/Consts.php":                                                           phpConsts,
+		"sdk/ruby/lib/datastar/consts.rb":                                                  rubyConsts,
```
bencroker: Maintain alphabetical order with the other SDKs, please.

sdk/ruby/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+## [Unreleased]
+
+## [0.1.0] - 2025-01-31
```
bencroker: I think you said you were going with 1.0.0-beta.1?

ismasan: Right. The Changelog file was auto-generated by my tooling. I'll remove it for now. Don't need the overhead until this is stable IMO.

BUILDING.md
ismasan: I'll take this change out of this PR for now. Might add as a separate PR.

sdk/ruby/README.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+### Sending updates to the browser
+
+There are two ways to use this gem in HTTP handlers:
+
+* One-off responses, where you want to send a single update down to the browser.
+* Streaming responses, where you want to send multiple updates down to the browser.
+
+#### One-off update:
```
bencroker, line 42: Why offer two modes? Why not just stream by default (like the PHP SDK does, for example)?

examples/ruby/progress.ru
bencroker, line 1: It seems like most of this example would be covered by the Hello World example, which is what we’d like all SDKs to implement.

sdk/ruby/README.md
ismasan: The "one-off" methods are just a shortcut for the streaming block with a single operation.

```ruby
def merge_fragments(fragments, options = BLANK_OPTIONS)
   stream do |sse|
     sse.merge_fragments(fragments, options)
   end
end
```

I do want the `stream` block because I want control over _when the stream ends_, and also because defining multiple stream blocks schedules the work in different threads transparently, which allows me to handle some framework-specific threading details under the hood, and to stream to the browser from concurrent threads, ex.

```ruby
# This stream runs in one thread
datastar.stream do |sse|
  100.times do |I|
    sleep 1
    sse.merge_fragments '<p id="slow">#{I}</p>`
  end
end

# This stream runs in another thread
datastar.stream do |sse|
  1000.times do |I|
    sleep 0.1
    sse.merge_fragments '<p id="fast">#{I}</p>`
  end
end
```

I expect most people to use the explicit `stream` block.

Finally, it's also in line with some existing streaming APIs, such as Sinatra's, which I use.

Also (and I haven't documented this), you can pass an interface to `#stream`, instead of a block.

```ruby
datastar.stream(SomeStreamObject.new)
```

I expect to use all this in my event-driven apps to allow subscribing to several server-side event sources and stream back to the client.

examples/ruby/progress.ru
ismasan: That wasn't clear to me. I'll add a hello world and remove this one.

sdk/ruby/README.md
bencroker: Ok, I’ll take your word for it.

examples/ruby/progress.ru
bencroker: Thanks!

ismasan: How do I code-generate that `hello-world.html` template into `examples/ruby/hello-world/`?

bencroker: https://github.com/starfederation/datastar/blob/d3b50bc3a99c1e527f422807038e05b805f6bae9/build/run.go#L145

ismasan: Done. 

https://github.com/user-attachments/assets/26c16cf1-96f6-4505-a5a6-f80cbfbe1b13

bencroker:
Thanks, you can mark as ready for review, whenever you feel it is.

bencroker:
Merged. Great work, thank you!!

PR #607, opened 2025-02-04
Add Accept: text/event-stream to bash tests and fix datastar-request header

Superpat:
Fix test headers

Overrides #570


bencroker:
Merged, thanks!

PR #608, opened 2025-02-04
Fix fragment morph

### LINKED ISSUE #588: Signals cleared when `datastar-merge-fragments` overwrites existing fragment with signals
take this example:
```
<div
  id="fragment"
  data-signals-foo="'bar'"
>
  <div data-text="$foo"></div>
  <button data-on-click="@get('/fragment')">reload</button>
</div>
```

`GET /fragment` merges the same fragment using `datastar-merge-fragments`.

## Expected

Pressing reload to just do nothing (since merged fragment sent back is exactly the same).

## Actual

Unfortunately instead the signals get cleared (foo is set to empty string and div disappears).

bencroker:
This fixes a bug discovered in Idiomorph that surfaced in 1.0.0-beta.3. It does so by pulling in [Idiomorph 0.4.0](https://github.com/bigskysoftware/idiomorph/releases/tag/v0.4.0). 

Fixes https://github.com/starfederation/datastar/issues/588.


bencroker:
It looks like the `removals` function is not being called in `mergeFragments.ts` before `applyAttributePlugin`. You can see how the number of requests increases each time you tab through in http://localhost:8080/examples/lazy_tabs

PR #609, opened 2025-02-05
switching tests back to one web server, many pages

zangster300:
tests do have a tendency to random fail due to flakiness, but big speed boost and aren't hanging

not sure what the best way is to handle the `markdownDir` variable in `router.go` needing to change for tests vs site deploy is at the moment


bencroker:
Merged, thanks!

PR #613, opened 2025-02-05
Fixed improper coordinates for using the clojure SDK

JeremS:
I mistakenly provided the wrong url in the docs, generating an error for users wanting to use the sdk. It is now fixed.


bencroker:
Merged, thanks!

PR #614, opened 2025-02-05
zig: change root to main

rofrol:
```
//! By convention, root.zig is the root source file when making a library. If
//! you are making an executable, the convention is to delete this file and
//! start with main.zig instead.
```

https://github.com/ziglang/zig/blob/ff551374a03c3310bdcbf6f5efa62b670037c3b0/lib/init/src/root.zig#L1


rofrol:
@jmstevers

rofrol:
Mistake. Sorry.

Issue #576, opened 2025-02-02
Python SDK: Add FastHTML response (and example)


axelknock:
FastHTML has an API very similar to FastAPI, except it uses python itself as the templating language through "FastTags", FT components. These will have to be converted to strings before being merged. I propose using `fasthtml.xml`'s `to_xml` and defaulting to accepting strings since `to_xml` can accept tuples and None as well. Let me know if I should make any other considerations @lllama

axelknock:
I forgot to link the docs: [FastHTML docs](https://docs.fastht.ml/). FastHTML is unique in that it was built with HTMX support for hypermedia-based approach, but from testing it's also amenable to being used with datastar. Almost a drop-in replacement but for the fact that endpoints have to put down SSE merge responses instead of HTML.

axelknock:
Merged #624

PR #624, opened 2025-02-06
feat(SDK): FastHTML SSE response and example (#576)

axelknock:
This pull request adds an override to the `merge_fragments` method when called with a DatastarFastHTMLResponse such that the fragments are first converted to valid HTML (from FastTags) before being merged.

It also includes an example of how to integrate Datastar with the Great Tables library, which could be useful in data science scenarios where users are interacting with tabular data.

The demonstration cannot be run until being merged into the datastar-py SDK library, but you can run another version with the same changes here using uv or any other tool that can run scripts with inline dependencies:
```sh
uv run https://gist.githubusercontent.com/axelknock/dac9d38e3f77edec774e50c7a080c0dd/raw/66b0a3a4571bf789e678bdd88ebd6178965f19bc/datastar-fasthtml-great-tables-demo.py
```


axelknock:
Sorry I forgot to tag @lllama

lllama:
LGTM :shipit:

axelknock:
Thanks Felix! @bencroker it also looks like the checks are now passing. Question: is the pypi release cut at the same time as the version releases? Does someone manually upload the package using twine or something?

lllama:
Hey @axelknock. Pypi releases are manual atm. Something that we should look into in the future.

bencroker:
Merged, thanks!

PR #647, opened 2025-02-11
Update sdks.md to point to correct examples for Typescript

alesvaupotic:
Link is 404 as all other languages live in /examples, while TS examples are in its own SDK folder. They should probably be moved with the others, but for now, at least links should point to where they are available currently.


bencroker:
Merged, thanks! 

Please think about how to improve this, @Superpat.

Issue #643, opened 2025-02-11
Stop catching error in `preact-core.ts`


bencroker:
We are currently catching errors in `preact-core.ts`, which is obfuscating the original error.

https://github.com/starfederation/datastar/blob/2b5e59c418141240d6d33bc59b0f3e21911a5eb7/library/src/vendored/preact-core.ts#L67-L74

This can be seen in the example at https://data-star.dev/examples/refs

```
Uncaught datastar internal error: BatchError, error
[...]

PR #651, opened 2025-02-12
643-Stop-catching-error-in-`preact-corets`

delaneyj:
We still want the error, just not to wrap it in another

### DIFF
--- a/library/src/vendored/preact-core.ts
+++ b/library/src/vendored/preact-core.ts
@@ -79,9 +79,7 @@ function endBatch() {
   batchIteration = 0
   batchDepth--
 
-  if (hasError) {
-    throw error
-  }
+  if (hasError) throw error
 }
 
 /**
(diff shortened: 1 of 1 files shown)


delaneyj:
Fixes #643

library/src/vendored/preact-core.ts
```diff
@@ -79,9 +79,7 @@ function endBatch() {
   batchIteration = 0
   batchDepth--
 
-  if (hasError) {
-    throw internalErr(from, 'BatchError, error', { error })
-  }
+  if (hasError) throw error
```
bencroker, line 82: Is there any reason to catch the error above (line 69), rather than simply allow it to be thrown?

delaneyj: cause you still need to unroll the dep before sending it to be in the right stack for the error

bencroker: I see, said the blind man.

Issue #645, opened 2025-02-11
Website Inline Validation Demo loses focus


jroberts-iii:
https://data-star.dev/examples/inline_validation
Go to that page, enter "test" into the Email Address, and then wait.

After 0.5 seconds, the input will lose focus.
It appears that merging the fragment returned by the @get causes the focus to be lost.

bencroker:
Thanks for the report! This likely has to do with the latest version of Idiomorph. Will look into it.

bencroker:
Released in [v1.0.0-beta.6](https://github.com/starfederation/datastar/releases/tag/v1.0.0-beta.6).

PR #653, opened 2025-02-12
Website Inline Validation Demo loses focus

delaneyj:
Hopefully fix a host of issues stemming from different semantics in idiomorph moving forward


delaneyj:
Fixes #645

PR #656, opened 2025-02-12
Bump version in Gemfile too

ismasan:
Apologies @bencroker I didn't commit the Gemfile.lock after the version bump before. Turns out it's necessary!


bencroker:
Merged!

PR #661, opened 2025-02-12
clarification on usage of kebab-case classname for data-class

phelex:
make notation of classnames with hyphens more clear in example 'kebab-case' as key


bencroker:
Thanks for the PR! I’ve chosen a more common class name for the clarification in https://github.com/starfederation/datastar/commit/8b1c46bd7b8fc549ef9923d5cb1b3e1ad3652b23#diff-449111daf7e50ee222c437eca294010bc5c0c119ef6543d4cf784791dc9f93bd.

PR #663, opened 2025-02-13
fix(watchers/mergeFragments): remove console.log

pblkt:


bencroker:
Thanks for the PR but we rarely accept library contributions that are not agreed upon beforehand. We’re aware of the stray `console.log` and it will be removed in the next release.

Issue #639, opened 2025-02-10
Lose and then regain window focus causes D* to trigger a request


nickchomey:
I dont know how to describe this better than this screen recording can - when the tab loses focus and then regains it, an sse GET fetch request is sent. Its not clear to me which fetch request is decided to be made though. 

It happens if you either change tabs, or otherwise cause the window to completely lose focus (split screen changing window focus doesnt do it)

I noticed this in my own app because one button creates users in IndexedDB and it was just continually growing. In fact, in my app, it triggers FOUR simultaneous requests when you regain focus, even though there's only one button with `@get` on the page. I can provide a screen recording of that if this one on the D* site isnt enough to diagnose the issue. 

https://github.com/user-attachments/assets/e1470c7b-d94b-4459-9650-f7cd52c4d937

delaneyj:
It's a feature... please read the reference before submitting bugs.  

https://data-star.dev/reference/action_plugins, specifically `openWhenHidden`

nickchomey:
Thank you and apologies for the trouble. 

Though, as it turns out, the real issue is actually a deeper user-error/bug in my code. I expect it'll be a very common error - by both new and experienced D* users - so I'll ponder on a way to perhaps change things to minimize that.

PR #667, opened 2025-02-13
Add a "keepAlive" connection option to the TS SDK

nickchomey:
@Superpat asked me to open this PR after I described this issue to him in private communications. 

I recently opened an erroneous bug report (#639) which stemmed from 
1) I wasn't aware of the `openWhenHidden` option in D*, and
2) More fundamentally, I had forgotten to close the SSE connection for an endpoint that should have just had a single response. 

This caused a SSE connection to open and stay open each time I clicked a button, then when I changed window focus, they'd all re-connect and make the same request (which was doing a DB mutation, among other things). 

Now, you might say that I'm just dumb, and you wouldn't be wrong. But I really have to figure that other people - even experienced D* devs - will fail to close streams, for a variety of good and bad reasons. 

So, given that D* wants to help people "fall into a pit of success", and that 0- and 1-response endpoints will likely vastly outnumber long-lived/infinite response endpoints, it seems to me that it would be prudent to have SSE connections default to closing automatically. A simple connection option (eg `keepAlive`) could then be passed in order to allow the connection to stay open indefinitely. 

This would be much safer (not unnecessarily drain battery nor make unwanted - and possibly infinite - DB mutations). Moreover, if a connection didn't stay open because you forgot to pass the `keepAlive` option, it would be immediately visible and easy to fix.

This PR adds such an option and it can be used like this `stream.mergeFragments(fragments, { keepAlive: true });`

I'm not entirely wedded to this particular implementation - it could have a different name or mechanism altogether. But I do think that *something* needs to be done to help people avoid such errors. Any alternative suggestions would be welcomed. 

I'm not sure how this would apply to the other SDKs, but I have to figure that the same issue probably occurs with them all.


Superpat:
You want to add the requirement to specify keepalive to every event ? I think that would differ too much from the sdk spec. But I'd be open to adding a keepalive option to the stream creation function (which is specific to the ts-sdk in any case)

You or @bencroker can close this PR, but I'll let you know when I've added the option.

nickchomey:
No, the current default is to keep the connection alive indefinitely. The intention of this PR is to make the default (without specifying the option) to NOT keeping the connection alive - just a single response, which is surely a considerably more common occurrence in most apps. 

If you want a long-lived connection to stream fragments to the browser, you could add the `keepAlive: true` option. Then close it manually when ready. 

This is what came to my mind to address the issue of people almost certainly leaving connections open by mistake - which is likely when they're new to D* and SSE, and surely still not-uncommon later on because errors happen. This seems like a sane, safe, low-friction solution. 

But no obligation to use this if you folks have a better approach to suggest.

Superpat:
Yeah but this way you need to add keepalive for every event that's not the last event. As you'll see in #698 I made it a stream wide decision.

nickchomey:
That sounds reasonable. Ill try to review and test that PR later today.

PR #681, opened 2025-02-20
Fix text overflow on buttons

akashgill3:
Fix button text overflow on small screens

- Add h-auto to allow buttons to grow with wrapped content
- Add consistent p-2 padding around button content
- Set min-width to prevent overly narrow buttons on mobile

Before this fix, button text would wrap outside the colored background on small screens. These changes ensure the background properly contains wrapped text while maintaining a responsive layout.

Tested on mobile and desktop viewports.
new_design


bencroker:
Perfect, thank you!

PR #683, opened 2025-02-21
docs: fix link to node.js example

jasalt:
[Typescript doc readme](https://github.com/starfederation/datastar/tree/main/sdk/typescript) has broken link to NodeJS example.
Correct file: https://github.com/starfederation/datastar/blob/main/sdk/typescript/examples/node.js


bencroker:
Thanks for the PR but we only accept PRs against the `develop` branch. I’ve fixed it in https://github.com/starfederation/datastar/commit/d245ab10a51bf3c96cf2a3227db33b8150cb6818.

Issue #602, opened 2025-02-04
Node sdk improvements


Superpat:
I need to do the same changes I did to the deno sdk, namely allow users to call the stream close themselves and better handling of aborts.

bencroker:
This is over 2 weeks old. Is there any movement on it?

Superpat:
Working on it. Just stuck recovering from my computer breaking. But trying to get operational asap.

PR #698, opened 2025-02-23
Typescript sdk improvements

Superpat:
- Brings node sdk in line with the web standard one
 - Update datastar version used in test servers
 - Simplify onError handler (dont think the error stream would of been very useful)
 - Add missing snippets
 - Fix test failures (was sending default value of reply)

@bencroker let me know if you have comments on the snippets

closes #667 and #602


bencroker (commented):
Looks great, thanks!

bencroker (approved):

sdk/typescript/src/web/serverSentEventGenerator.ts
```diff
@@ -32,44 +32,48 @@ export class ServerSentEventGenerator extends AbstractSSEGenerator {
...
     const readableStream = new ReadableStream({
       async start(controller) {
         const generator = new ServerSentEventGenerator(controller);
 
         try {
-          const startedStream = onStart(generator);
-          if (startedStream instanceof Promise) await startedStream;
+          const stream = onStart(generator);
+          if (stream instanceof Promise) await stream;
+          if (options?.keepalive) {
```
nickchomey, line 61: @Superpat should this be `!options?.keepalive`, like in the node generator, so that it only stays upon when you pass keepalive = true?

Superpat: argh you're right

PR #705, opened 2025-02-25
Fix sdk npm package

Superpat:
I think this should fix the problem people are having with the npm package.


bencroker (approved):

bencroker:
Merged, thanks!

PR #711, opened 2025-02-26
Fix merge fragments

### LINKED ISSUE #709: data-show or data-on-click issue after upgrading from beta7 to beta8.
Initial page load contains:
```html
<div data-signals-showlogreg="false">
...
  <a href="#" data-on-click={ datastar.GetSSE("/login") }>Login</a>
   or
  <a href="#" data-on-click={ datastar.GetSSE("/register") }>Register</a>
  <div id="logreg" data-show="$showlogreg"></div>
...
```

Both templ fragments return:
```html
<div id="logreg" data-show="$showlogreg" data-on-load="$showlogreg=true">
  <a href="#" data-on-click="$showlogreg=false">x</a>
  ...
```

In 1.0.0-beta.7: behaves as expected: clicking x will hide the div no matter how many times you interact with the login/register/x clicks.
In 1.0.0-beta.8: correctly hides the div only the first time the x is clicked.  So: Click login, click x, click login again, click x.  This 2nd call to data-on-click to set showlogreg=false fails.

go.mod:
```
github.com/starfederation/datastar v1.0.0-beta.8
```

in the templ page:
```html
<script type="module" src="https://cdn.jsdelivr.net/gh/starfederation/datastar@v1.0.0-beta.8/bundles/datastar.js"></script>
```

### LINKED ISSUE #710: Issue with focus being lost on input beta7->beta8
This snippet below exposes a behavior change to the way inputs and merge fragments interact. In version beta7 and below, the focus would stay on the text input when a fragment came in. However in beta8, now it appears to lose focus on merge fragment. This is true for when a signal is bound to the input.

Snippet:
```
<body class="bg-gray-100 h-screen" id="body" data-on-load="@get(`/overview/sse`, {})">
    <div class="container mx-auto p-4" id="testing">
<div data-signals__ifmissing="{'test' : 'test'}" data-on-signals-change-test__debounce.20ms="@get(`/overview/search`, {})" id="datastar58uydy"><input data-bind-test  id="datastar8m3obv"></div>
</div>
</body>
```
**Backend:**
The backend can just send a merge fragment for this html.

Maybe I'm missing something here, but I tried to boil the problem down to the simplest example.

bencroker:
This PR fixes #709 in which `data-signals` was being reapplied each time any attribute changed on an element, caused by the fact that it doesn’t return a cleanup function so a hash was never created for it – a real headfuck to debug.

The fix includes storing an empty cleanup function for plugins that don’t return one. It also rehashes the cleanup functions for each element being merged to ensure that plugins are cleaned up and reapplied after merging.

I’ve also restored the assigning of unique element IDs to fragments being merged, to allow focus to be restored to focusable elements, which fixes #710.

@delaneyj note that I changed the `removals` key from an element to an element ID to allow for rehashing in the `mergeFragments` plugin.


kaldyr:
Fixes #709

library/src/engine/engine.ts
```diff
@@ -25,23 +25,25 @@ const plugins: AttributePlugin[] = []
 const actions: ActionPlugins = {}
 const watchers: WatcherPlugin[] = []
 
+// Map of cleanup functions by element ID, keyed by a dataset key-value hash
+const removals = new Map<string, Map<number, OnRemovalFn>>()
```
bencroker, line 29: @delaneyj note that I changed the key from an element to an element ID to allow for rehashing in the `mergeFramgents` plugin.

delaneyj (approved):

Issue #707, opened 2025-02-25
Inconsistent SSE message parsing in Beta 8 (datastar bug and/or python SDK bug)


axelknock:
In Beta 7 the following `mergeMode` directive works, in Beta 8 it doesn't:
```
event:·datastar-merge-fragments
retry:·1000
data:·mergeMode··after
data:·selector··#notifications
data:·useViewTransition··false
data:·fragments··<div id="notification-nb9kpwm3" class="notification info">test</div>
```
Note that `·` represents whitespace.

In either version the `fragments` directive does work, which is what makes me think that the parsing is inconsistently stripping whitespace.

In the python SDK there are some extraneous spaces added in `sse.py` because the `MergeModeDatalineLiteral` const in `const.py` already includes a whitespace. Whether that is necessarily a bug depends on whether whitespace should generally be stripped from the directive values. If we're liberal with whitespace than as long as it's on the same line as the directive then it shouldn't fail.

I can see why the `fragments` directive is treated differently, but I do not think the behavior should be inconsistent between directives.

If the amount of whitespace does matter then the `fragments` directive should also fail but doesn't, ie they should all fail or none of them should fail. I'll look into whether this is in the HTML spec for guidance.

axelknock:
I could find no guidance on parsing inside the line after the colon, [only at and before](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation).

I tend to think that we should be strict about whitespace and limit it to a single space in which case `fragments` should be brought into compliance with the rest of the directives, allowing for only single spaces (and fixing this in the python SDK and any other ones affected).

bencroker:
I’m not seeing an extraneous space in the constant. Can you please ensure you’re running the latest version of the SDK?

https://github.com/starfederation/datastar/blob/57a0f468ac89e5b55593a292b5566404e4e5c8e2/sdk/python/src/datastar_py/consts.py#L74

axelknock:
You are right the constant no longer includes that space and will interpolate correctly. The latest release is still `0.4.2` so the version number should probably be changed for the next pypi release.

bencroker:
Can you please bump the version, @lllama, or remind me what I need to do if it requires some action from my side?

lllama:
[...]

PR #714, opened 2025-02-27
Python SDK fix: fasthtml correct imports and raise version (#576)

axelknock:
Corrects imports (they were being shadowed earlier so I got the wrong names) and raised version which should have been raised before.


axelknock:
Should also mention [#707](https://github.com/starfederation/datastar/issues/707#issuecomment-2683832434):
> Can you please bump the version, [@lllama](https://github.com/lllama), or remind me what I need to do if it requires some action from my side?
- @bencroker

I'll look at pypi package deployment for the python SDK.

bencroker:
Review please, @lllama.

sdk/python/pyproject.toml
```diff
@@ -37,11 +37,12 @@ build-backend = "hatchling.build"
...
     "django>=4.2.17",
     "fastapi[standard]>=0.115.4",
-    "fasthtml>=0.12.0",
+    "python-fasthtml==0.12.0",
     "flask[async]>=3.0.3",
     "quart>=0.19.9",
     "sanic>=24.6.0",
     "uvicorn>=0.32.1",
+    "bump-my-version>=0.32.0",
 ]
```
dalito: Above the version is declared as dynamic. But then you use hard-coded version below and add a tool here. So is the idea to use dynamic versioning or not?

sdk/python/pyproject.toml
```diff
@@ -37,11 +37,12 @@ build-backend = "hatchling.build"
 dev = [
     "django>=4.2.17",
     "fastapi[standard]>=0.115.4",
-    "fasthtml>=0.12.0",
+    "python-fasthtml==0.12.0",
```
dalito: Is a hard pin for an SDK is a good idea? Is it really required for python-fasthtml?

sdk/python/pyproject.toml
axelknock: I added this because @lllama included a declaration for the bumpmyversion tool. When I tried to use the tool it failed due to the repo tags interfering with the SDK's versioning (and I could not figure out how to disregard the tag version), so it might be worth discarding both this and the tool configuration, but as far as I can tell @lllama currently manually distributes the SDK to pypi so I defer to him.

sdk/python/pyproject.toml
axelknock: True.

sdk/python/pyproject.toml
lllama: So I use `uv` for building and publishing the sdk to pypi. I think it'll probably be best to use `uv` for running `bump-my-version` as well.

axelknock: @lllama I can't manage to run bump-my-version inside the datastar repo without the repo's tags interfering with the command. I don't see any obvious way to ignore the tags. How do you manage to run the command? Do you have the python SDK outside the repo?

axelknock:
@bencroker @lllama said on discord that there are too many unrelated changes in this PR. I will cut it back tomorrow and comment when it's done.

axelknock:
Apologies for not getting this done sooner @bencroker, @lllama. I've cut everything but the fix and dynamic fasthtml version.

Kvit:
@lllama  all still works in alex's branch

bencroker:
Review please, @lllama.

axelknock:
I made a change suggested by @gazpachoking on discord to put the fasthtml requirement behind an env variable since its minimum python version is higher than datastar-py. I confirmed it works with uv sync. Let me know if I should make any other changes @lllama

gazpachoking:
This is still valid isn't it @axelknock?

axelknock:
@gazpachoking I screwed up when making this PR by using my `develop` branch and I don't want it tied to this PR anymore. I'll open a new PR for the fix.

PR #723, opened 2025-03-01
docs fixing

gavriguy:
change `While you want to long expressions` to `While you don't want to long expressions`


bencroker:
Thanks for the PR! I opted for better wording in https://github.com/starfederation/datastar/commit/d8b22b58fa8969ebc012ca73f130de24489e456f.

PR #730, opened 2025-03-02
Add user expression args

bencroker:
This PR replaces the `ctx` argument passed into user expressions with `el` and `signals`. By doing so, we reduce the surface area of the Datastar expression API, making it easier to document and maintain long-term.

Before:

```html
<div data-on-click="ctx.el.value = 0"
     data-text="ctx.signals.JSON()"
></div>
```

After:

```html
<div data-on-click="el.value = 0"
     data-text="signals.JSON()"
></div>
```

And since signals are already JSON encoded when output, this can be written as:

```html
<div data-on-click="el.value = 0"
     data-text="signals"
></div>
```


bencroker:
We decided against doing this. If we’re going to inject a variable into the expression, it might as well just be `ctx`.

PR #738, opened 2025-03-05
POC: brotli and gzip encoding option for zig sdk

0x546F6D:
In their last podcast on youtube, Delaney and Ben talked about brotli encoding being worked on for the clojure sdk. I too am interested in a similar feature for the zig sdk, so I created this draft as a proof of concept.
While I was at it, I also included the gzip encoding.

With encoding, you need to concatenate all the merge/remove/executes commands into 1 single response that is then encoded before being sent back to the browser as 1 chunk of data.
This disqualifies long lasting sse connections, but it might still be useful when you know that you will be sending a big fragment of data for some specific requests.

My current workaround is to add a "pub fn sendEncoded()" to the API, which should be called after all the mergeFragments/Signals/etc.. if you chose to use encoding, resulting in this code:
```zig
const ds = @import("datastar");
const dsz = ds.httpz;

// Send response with Brotli encoding:
    var sse = try dsz.ServerSentEventGenerator.init(res, .{ .encoding = .br }); // for Brotli, another choice is .gzip
    defer sse.sendEncoded() catch {};
    try sse.mergeFragments(very_big_fragment, .{ .selector = "#overlay", .merge_mode = .inner });
    try sse.mergeSignals(.{ ._overlay_display = true }, .{});
    // Or `try sse.sendEncoded();` at the end instead of using `defer`

// Send response without encoding:
// no change compared to before except for the .{} option required by init():
    var sse = try dsz.ServerSentEventGenerator.init(res, .{}); // encoding is null by default
    try sse.mergeFragments(normal_fragment, .{ .selector = "#overlay", .merge_mode = .inner });
    try sse.mergeSignals(.{ ._overlay_display = true }, .{});
```

### DIFF
--- a/sdk/zig/src/ServerSentEventGenerator.zig
+++ b/sdk/zig/src/ServerSentEventGenerator.zig
@@ -1,10 +1,21 @@
 const std = @import("std");
 const consts = @import("consts.zig");
+const Brotli = @import("brotli");
+const br = Brotli.init(Brotli.Settings{});
 
 const default_execute_script_attributes: []const []const u8 = &[_][]const u8{consts.default_execute_script_attributes};
 
 allocator: std.mem.Allocator,
 writer: std.net.Stream.Writer,
+encoding: ?consts.Encoding = null,
+data: std.ArrayList(u8) = undefined,
+
+pub const InitOptions = struct {
+    /// If an `encoding` is selected, all the merge/remove/execute commands will be concatenated,
+    /// encoded, and then sent to the client.
+    /// Using 'encoding' requires the use of the function `sse.sendEncoded()` after all those commands
+    encoding: ?consts.Encoding = null,
+};
 
 pub const ExecuteScriptOptions = struct {
     /// `event_id` can be used by the backend to replay events.
@@ -77,32 +88,78 @@ pub const RemoveSignalsOptions = struct {
     retry_duration: u32 = consts.default_sse_retry_duration,
 };
 
+pub const SendOptions = struct {
+    /// `event_id` can be used by the backend to replay events.
+    /// This is part of the SSE spec and is used to tell the browser how to handle the event.
+    /// For more details see https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#id
+    event_id: ?[]const u8 = null,
+    /// `retry_duration` is part of the SSE spec and is used to tell the browser how long to wait before reconnecting if the connection is lost.
+    /// For more details see https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#retry
+    retry_duration: u32 = consts.default_sse_retry_duration,
+};
+
 fn send(
     self: *@This(),
     event: consts.EventType,
     data: []const u8,
-    options: struct {
-        event_id: ?[]const u8 = null,
-        retry_duration: u32 = consts.default_sse_retry_duration,
-    },
+    options: SendOptions,
 ) !void {
-    try self.writer.print("event: {}\n", .{event});
+    if (self.encoding) |_|
+        try write(
+            self.data.writer(),
+            event,
+            data,
+            options,
+        )
+    else
+        try write(
+            self.writer,
+            event,
+            data,
+            options,
+        );
+}
+
+/// Write data to sse stream or self.data according to encoding
+fn write(
+    writer: anytype,
+    event: consts.EventType,
+    data: []const u8,
+    options: SendOptions,
+) !void {
+    try writer.print("event: {}\n", .{event});
 
     if (options.event_id) |id| {
-        try self.writer.print("id: {s}\n", .{id});
+        try writer.print("id: {s}\n", .{id});
     }
 
     if (options.retry_duration != consts.default_sse_retry_duration) {
-        try self.writer.print("retry: {d}\n", .{options.retry_duration});
+        try writer.print("retry: {d}\n", .{options.retry_duration});
     }
 
     var iter = std.mem.splitScalar(u8, data, '\n');
     while (iter.next()) |line| {
         if (line.len == 0) continue;
-        try self.writer.print("data: {s}\n", .{line});
+        try writer.print("data: {s}\n", .{line});
     }
 
-    try self.writer.writeAll("\n\n");
+    try writer.writeAll("\n\n");
+}
+
+/// Send encoded msg to the browser
+pub fn sendEncoded(
+    self: *@This(),
+) !void {
+    if (self.encoding) |encoding| switch (encoding) {
+        .br => {
+            const encoded = try br.encode(self.allocator, try self.data.toOwnedSlice());
+            try self.writer.writeAll(encoded);
+        },
+        .gzip => {
+            var fbs = std.io.fixedBufferStream(try self.data.toOwnedSlice());
+            try std.compress.gzip.compress(fbs.reader(), self.writer, .{});
+        },
+    };
 }
 
 /// `executeScript` executes JavaScript in the browser
--- a/sdk/zig/src/ServerSentEventGenerator.zig
+++ b/sdk/zig/src/ServerSentEventGenerator.zig
@@ -7,7 +7,8 @@ const default_execute_script_attributes: []const []const u8 = &[_][]const u8{con
 
 allocator: std.mem.Allocator,
 writer: std.net.Stream.Writer,
-encoding: ?consts.Encoding = null,
+res: *@import("httpz").Response,
+options: InitOptions,
 data: std.ArrayList(u8) = undefined,
 
 pub const InitOptions = struct {
@@ -15,6 +16,9 @@ pub const InitOptions = struct {
     /// encoded, and then sent to the client.
     /// Using 'encoding' requires the use of the function `sse.sendEncoded()` after all those commands
     encoding: ?consts.Encoding = null,
+    /// Minimum size for the datastar sse response to trigger encoding.
+    /// Responses with size below this treshold will be sent as plain text without being encoded
+    enc_min_size: u16 = 256,
 };
 
 pub const ExecuteScriptOptions = struct {
@@ -104,7 +108,7 @@ fn send(
     data: []const u8,
     options: SendOptions,
 ) !void {
-    if (self.encoding) |_|
+    if (self.options.encoding) |_|
         try write(
             self.data.writer(),
             event,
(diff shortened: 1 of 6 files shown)


bencroker:
Paging @jmstevers to take a look.

ghost:
This is a good idea, although I think this can be done without `sendEncoded`. If you write to a buffer by default and optionally encode it then send it to the writer it should work and be a lot more ergonomic. Also if you were to set the http connection handover to be `.keepalive` (maybe via the `InitOptions` struct) you can do long lived connections. I would do this myself but zig master doesn't compile on my machine right now haha. I don't like this relying on your personal brotli library, but if you plan to maintain it, its not really that big of an issue because the codebase is fairly small.

edit: There should also be a way of setting a minimum size for the buffer before its compressed (e.g. only compress above 256 bytes).

0x546F6D:
> This is a good idea

I am glad to hear that since this was the main goal of this PR: have a proof of concept out in the open to discuss the usefulness of implementing this feature in the zig sdk.

> although I think this can be done without `sendEncoded`. If you write to a buffer by default and optionally encode it then send it to the writer it should work and be a lot more ergonomic. Also if you were to set the http connection handover to be `.keepalive` (maybe via the `InitOptions` struct) you can do long lived connections.

You lost me :D. I am not sure that I have near enough knowledge in tcp/stream/encoding nor zig to be able to code that but I'll wait until you take over the implementation of this feature since as they say, ~a picture is~ a few lines of code are worth a 1000 words :)

> edit: There should also be a way of setting a minimum size for the buffer before its compressed (e.g. only compress above 256 bytes).

That I can understand and gave it a try.. However, as seen in the commit I just pushed, I couldn't figure out how to do it without having to pass the whole httpz.response to the ServerSentEventGenerator, since I am currently stuck with the sendEncoded() function. This now makes passing the res.arena and the std.net.Stream.Writer redundant.

> I would do this myself but zig master doesn't compile on my machine right now haha.

No worries, there is no rush. I tried to follow your style of coding (and learn from it for my own project :)) So I hope that this draft can still be of use as a basis.

> I don't like this relying on your personal brotli library, but if you plan to maintain it, its not really that big of an issue because the codebase is fairly small.

I completely understand and I hope that this is just temporary until brotli is added to the std. There is interest to request it ([Add Brotli decompressor](https://github.com/ziglang/zig/issues/22445)) but it is currently in the unplanned milestone.

In the meantime, I wanted to use Brotli as well as to learn how to package a C project for the Zig build system.. 2 birds 1 stone

I followed the guidelines of [All Your Codebase](https://github.com/allyourcodebase) and created 2 repositories:
- 1 repo that is limited to only packaging brotli for the zig build system
- Another 1 that acts as a minimal zig wrapper around the encode/decode C functions using the 1st repo as a dependency.

I try to maintain those 2 repos as best as I can since I also use them in my own project, like modifying their build.zig.zon this week after the [implement new package hash format](https://github.com/ziglang/zig/pull/22994) PR got released, or update the google/brotli dependency when there is a new commit on their side.

However, the code contained in those 2 repos can easily be combined and added to this zig sdk. This would make the build.zig a bit more complicated but it would also mean that this zig sdk would depend directly on [google/brotli](https://github.com/google/brotli) without any middlemen like my repos.

I used my own brotli library in this PR because that's what I am more comfortable with but I know of 2 other options:
- [hexops/brotli](https://github.com/hexops/brotli), which uses a bash script to update its repo.
- [bun/src/brotli.zig](https://github.com/oven-sh/bun/tree/main/src/brotli.zig), but I'll admit that this implementation went a bit over my head :)

Let me know if I can be of any help.

sdk/zig/src/httpz/ServerSentEventGenerator.zig
```diff
@@ -3,21 +3,25 @@ const config = @import("config");
...
+pub fn init(res: *httpz.Response, options: ServerSentEventGenerator.InitOptions) !ServerSentEventGenerator {
     res.content_type = .EVENTS;
     res.header("Cache-Control", "no-cache");
 
     if (config.http1) {
         res.header("Connection", "keep-alive");
     }
 
-    try res.write();
+    if (options.encoding) |_| {} else try res.write();
```
ghost, line 14: `if (options.encoding == null) try res.write();`

sdk/zig/src/consts.zig
ghost, line 1: Don't modify this file, its auto generated by the D* build. Add `Encoding` to `root.zig` or `ServerSentEventGenerator.zig`

sdk/zig/src/ServerSentEventGenerator.zig
```diff
@@ -77,32 +92,88 @@ pub const RemoveSignalsOptions = struct {
...
+        try write(
+            self.writer,
+            event,
+            data,
+            options,
+        );
+}
+
+/// Write data to sse stream or self.data according to encoding
+fn write(
```
ghost, line 128: Put `write` back into send and just do `const writer = if (self.options.encoding) |_| self.data.writer() else self.writer;`

sdk/zig/src/ServerSentEventGenerator.zig
```diff
@@ -77,32 +92,88 @@ pub const RemoveSignalsOptions = struct {
...
 
-    try self.writer.writeAll("\n\n");
+    try writer.writeAll("\n\n");
+}
+
+/// Send encoded msg to the browser
+pub fn sendEncoded(
+    self: *@This(),
+) !void {
+    if (self.options.encoding) |encoding| {
```
ghost, line 157: Use an if guard instead of nested ifs. `const encoding = if (self.options.encoding) |enc| enc else return;`

sdk/zig/src/httpz/ServerSentEventGenerator.zig
```diff
@@ -3,21 +3,25 @@ const config = @import("config");
...
     const conn = res.conn;
     conn.handover = .close;
 
     return .{
         .allocator = res.arena,
         .writer = conn.stream.writer(),
+        .res = res,
+        // .encoding = options.encoding,
+        .options = options,
+        .data = if (options.encoding) |_| std.ArrayList(u8).init(res.arena) else undefined,
```
ghost, line 25: Just pass in `res` and `options` and create everything inside `ServerSentEventGenerator`

ghost (changes requested):

ghost:
>You lost me :D. I am not sure that I have near enough knowledge in tcp/stream/encoding nor zig to be able to code that but I'll wait until you take over the implementation of this feature since as they say, a picture is a few lines of code are worth a 1000 words :)

Just default to writing to the `data` buffer on the `ServerSentEventGenerator`, then optionally compress `data` (sending the `Content-Encoding` header if you do), then do `try res.conn.stream.writer().write(data);`. For long lived connections you'll need to ignore `enc_min_size`.

sdk/zig/src/ServerSentEventGenerator.zig
0x546F6D: I tried that 1st, zig complained about different types of writers. Hence why I decided to create the write(writer: anytype, ..) as workaround.

sdk/zig/src/consts.zig
0x546F6D: That's where it was at 1st, but then noticed that all the enums where in const.zig and moved it there. I did not realized that this file was auto-generated. Now putting it back in ServerSentEventGenerator.zig.

bencroker:
I do question whether this belongs in the SDK. If it is going to add bloat and maintenance burden, please consider whether it makes more sense to extract to an external package (similar to how we have packages that build on top of the PHP SDK). Any thoughts, @jmstevers?

0x546F6D:
Leaving only gzip, which is part of zig standard library would remove the issue about the brotli dependency's maintenance.

However, maintenance is pretty much a Zig problem, leaving people who follow the master branch with breaking changes every few months :)
For example, after the breaking changes of last week, @jmstevers created a PR on the 2nd of March to reflect those changes, http.zig and tokamak, which this sdk is based on, got updated on the 2nd and 4th, and I too updated my repos on the 3rd.

Too mitigate this issue, the zig package manager works with specific commits, while many zig repos create stable branches for different versions of zig. Now that zig 0.14 just got released, starfederation/datastar-zig could also maybe have a stable zig-0.14 branch that depends on a stable zig-0.14 branch/commit of its dependencies?

sdk/zig/src/ServerSentEventGenerator.zig
ghost: `writer.any()` should give a generic writer type

ghost:
> I do question whether this belongs in the SDK. If it is going to add bloat and maintenance burden, please consider whether it makes more sense to extract to an external package (similar to how we have packages that build on top of the PHP SDK). Any thoughts, @jmstevers?

True, the compression of an sse stream should not be d* sdk specific. Maybe the effort put into this should be repurposed into an http.zig middleware.

sdk/zig/src/ServerSentEventGenerator.zig
0x546F6D: Thanks, that worked 👍

sdk/zig/src/httpz/ServerSentEventGenerator.zig
0x546F6D: I have a question.
I have not used Tokamak so far, but since it is based on httpz, its request/response are defined in its main.zig as:
```zig
pub const Request = httpz.Request;
pub const Response = httpz.Response;
```
This is probably why zig is currently not complaining when I define `res` as an `httpz.Response` in the main ServerSentEventGenerator.zig.
```zig
res: *@import("httpz").Response,
```
Unless Tokamak decides to change its `req/res` in the future, wouldn't it be possible to merge the readSignals()/init() functions of the Tokamak/httpz folders into the main ServerSentEventGenerator?

On the other hand, if Tokamak does decide to use custom `req/res` in the future, then passing the whole `*tk.res/*httpz.res` instead of just `res.arena/conn.stream.writer()` to the main ServerSentEventGenerator.zig would not compile anymore as they would be different types.

Is there a way to pass only the `res.header/write` functions to ServerSentEventGenerator.zig? I tried using the following declaration in the ServerSentEventGenerator.zig but was unsuccessful because `header` is not a field of `httpz.Response`:
```zig
header: *const fn (*anyopaque, []const u8, []const u8) void,
```

0x546F6D:
> > I do question whether this belongs in the SDK. If it is going to add bloat and maintenance burden, please consider whether it makes more sense to extract to an external package (similar to how we have packages that build on top of the PHP SDK). Any thoughts, @jmstevers?
> 
> True, the compression of an sse stream should not be d* sdk specific. Maybe the effort put into this should be repurposed into an http.zig middleware.

I will study http.zig middleware, and try go that road 👍 Thanks a lot for the valuable inputs.

PR #747, opened 2025-03-07
Clojure SDK - Ability to gzip sse streams - improvements - new how to code snippet

JeremS:
This is a decently sized release. The internals of the SDK have been remade to cleanly add the ability to compress a SSE event stream. There are a number of other additions:
- new malli schemas
- new / improved / fixed documentation
- new examples using compression
- overhaul of the ring adapter making it behave more like the Http-kit one
- new tests to go with the new functionality

More details are in the SDK changelog.

The hope with this release is that the Clojure SDK shouldn't need any more features.

### DIFF
--- a/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
+++ b/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
@@ -10,6 +10,8 @@
 
 
 (def basic-profile
+  "Basic write profile using temporary [[StringBuilder]]s, no output stream and
+  no compression."
   {ac/write! (ac/->build-event-str)})
 
 
--- a/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
+++ b/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
@@ -1,42 +1,95 @@
 (ns starfederation.datastar.clojure.adapter.http-kit.impl
   (:require
-    [starfederation.datastar.clojure.api.sse :as sse]
+    [starfederation.datastar.clojure.adapter.common :as ac]
     [starfederation.datastar.clojure.protocols :as p]
     [starfederation.datastar.clojure.utils :as u]
     [org.httpkit.server :as hk-server])
   (:import
     [java.util.concurrent.locks ReentrantLock]
-    [java.io Closeable]))
+    [java.io ByteArrayOutputStream Closeable IOException]))
 
 
 ;; -----------------------------------------------------------------------------
-;; HTTP-KIT sse gen
+;; Sending the headers
 ;; -----------------------------------------------------------------------------
-(defrecord HK-gen [ch lock]
+(defn send-base-sse-response!
+  "Send the response headers, this should be done as soon as the conneciton is
+  open."
+  [ch req {:keys [status] :as opts}]
+  (hk-server/send! ch
+                   {:status (or status 200)
+                    :headers (ac/headers req opts)}
+                   false))
+
+;; -----------------------------------------------------------------------------
+;; Sending events machinery
+;; -----------------------------------------------------------------------------
+(defn ->send-simple [ch opts]
+  (let [buffer-size (or (ac/buffer-size opts)
+                        ac/default-write-buffer-size)]
+    (fn
+      ([])
+      ([event-type data-lines opts]
+       (let [event (ac/build-event-str buffer-size event-type data-lines opts)]
+         (hk-server/send! ch event false))))))
+
+
+(defn flush-baos! [^ByteArrayOutputStream baos ch]
+  (let [msg (.toByteArray baos)]
+    (.reset baos)
+    (hk-server/send! ch msg false)))
+
+
+(defn ->send-gzip [ch opts]
+  (let [^ByteArrayOutputStream baos (ByteArrayOutputStream.)
+        {:keys [writer write!]} (ac/->write-machinery baos opts)]
+    (fn
+      ([]
+       ;; Close the writer first to finish the gzip process
+       (.close ^Closeable writer)
+       ;; Flush towards SSE out
+       (flush-baos! baos ch))
+      ([event-type data-lines opts]
+       (write! writer event-type data-lines opts)
+       (ac/flush writer)
+       (flush-baos! baos ch)))))
+
+
+(defn ->send! [ch opts]
+  (if (or (ac/gzip? opts)
+          (ac/hold-write-buff? opts))
+    (->send-gzip ch opts)
+    (->send-simple ch opts)))
+
+;; -----------------------------------------------------------------------------
+;; SSE gen
+;; -----------------------------------------------------------------------------
+(deftype SSEGenerator [ch lock send!]
   p/SSEGenerator
-  (send-event! [_ event-type data-lines opts]
-    (let [event-str (-> (StringBuilder.)
-                        (sse/write-event! event-type data-lines opts)
-                        str)]
-      (u/lock! lock
-        (hk-server/send! ch event-str false))))
+  (send-event! [this event-type data-lines opts]
+    (u/lock! lock
+      (try
+        (send! event-type data-lines opts)
+        (catch IOException _ false)
+        (catch Exception e
+          (throw (ex-info "Error sending SSE event"
+                          {:sse-gen this
+                           :event-type event-type
+                           :data-lines data-lines
+                           :opts opts}
+                          e))))))
+
+  (get-lock [_] lock)
 
   (close-sse! [_]
-    (u/lock! lock
-      (hk-server/close ch)))
+    (hk-server/close ch))
+
+  (sse-gen? [_] true)
 
   Closeable
   (close [this]
     (p/close-sse! this)))
 
 
-(defn ->sse-gen [ch]
-  (HK-gen. ch (ReentrantLock.)))
-
-
-(defn send-base-sse-response! [ch req {:keys [status headers]}]
-  (hk-server/send! ch
-                   {:status (or status 200)
-                    :headers (merge headers (sse/headers req))}
-                   false))
-
+(defn ->sse-gen [ch send!]
+  (SSEGenerator. ch (ReentrantLock.) send!))
--- a/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
+++ b/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
@@ -70,7 +70,9 @@
     (u/lock! lock
       (try
         (send! event-type data-lines opts)
-        (catch IOException _ false)
+        (catch IOException _
+          (p/close-sse! this)
+          false)
         (catch Exception e
           (throw (ex-info "Error sending SSE event"
                           {:sse-gen this
--- a/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
+++ b/sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
@@ -9,6 +9,10 @@
     [java.io ByteArrayOutputStream Closeable IOException]))
 
 
+(def basic-profile
+  {ac/write! (ac/->build-event-str)})
+
+
 ;; -----------------------------------------------------------------------------
 ;; Sending the headers
 ;; -----------------------------------------------------------------------------
@@ -24,13 +28,12 @@
 ;; -----------------------------------------------------------------------------
 ;; Sending events machinery
 ;; -----------------------------------------------------------------------------
-(defn ->send-simple [ch opts]
-  (let [buffer-size (or (ac/buffer-size opts)
-                        ac/default-write-buffer-size)]
+(defn ->send-simple [ch write-profile]
+  (let [write! (ac/write! write-profile)]
     (fn
       ([])
       ([event-type data-lines opts]
-       (let [event (ac/build-event-str buffer-size event-type data-lines opts)]
+       (let [event (write! event-type data-lines opts)]
          (hk-server/send! ch event false))))))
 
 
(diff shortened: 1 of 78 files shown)


sdk/clojure/src/dev/examples/snippets/load_more.clj
bencroker, line 1: Is there a reason this file has to be so long? As a snippet in a how-to guide, it should be as concise as possible.

sdk/clojure/src/dev/examples/snippets/polling1.clj
bencroker, line 1: Can you explain why `#_{:clj-kondo/ignore true}` is found all over the place?

sdk/clojure/src/dev/examples/snippets/load_more.clj
JeremS: This file contains a version of the snippet I use to test at the repl. The actual snippet is in the directory with the other snippets and is ~30 lines.

sdk/clojure/src/dev/examples/snippets/polling1.clj
JeremS: Clj-kondo is a linter / static analyser for Clojure code that comes with clojure-lsp. In some cases it generates warnings that aren't relevant because it doesn't understand the code properly. This form `#_{:clj-kondo/ignore true}` allows me to suppress these warnings. 

In this particular instance, I have 2 forms requiring namespaces in the same file:
- one at the top to require what I actually need to test the code at the repl
- one wrapped in a comment containing the require form I copy paste in the actual snippet file.

In this case Clj-kondo doesn't take into account that the second require form is wrapped in a comment. It generates a warning telling me I require the same namespace twice even though I don't.

I use this pattern for snippet test files and it generates a lot of irrelevant warnings. There are other cases where Clj-kondo can't analyze code properly, often when using macros. I was starting to get too much of these warnings, so in this release I suppressed the false positives.

Note than I only do that for code in the `dev` namespace which contains examples and experiments I use while working on the SDK. This is not code that actually goes into the SDK proper.

sdk/clojure/src/dev/examples/snippets/load_more.clj
bencroker: Ah yes, different file :)

sdk/clojure/src/dev/examples/snippets/polling1.clj
bencroker: Thanks for the explanation.

bencroker (approved):

bencroker:
@andersmurphy since this is a big PR, could you please review?

sdk/clojure/adapter-http-kit/src/main/starfederation/datastar/clojure/adapter/http_kit/impl.clj
```diff
@@ -1,42 +1,97 @@
...
+      ([event-type data-lines opts]
+       (write! writer event-type data-lines opts)
+       (ac/flush writer)
+       (flush-baos! baos ch)))))
+
+
+(defn ->send! [ch opts]
+  (if (or (ac/gzip? opts)
+          (ac/hold-write-buff? opts))
+    (->send-gzip ch opts)
```
Ramblurr: What if a dev wants to use brotli or something besides gzip? Should the choice of compression function be injectable somehow? (my first thought on seeing this is that I would want to switch to brotli)

andersmurphy: @JeremS I do wonder if passing an "compression" `OutputStream` might be cleaner, if it can be done. Rather than special casing gzip. Then the various buffer options become compression-buffer options. As with both brotli and gzip you can configure the buffer size.

JeremS: Ok so I have been thinking. It's true that plugable compression would be better. I have spent some time designing a different api, I'd like your feed back on this.

The idea is to let the user in control of the OutputStream and the way it is wrapped (for adding compression, buffering ...)

Instead of the `gzip` option you would pass a `write-profile` one. This option would be a map of 3 keys:
- a wrap function from an output stream to a wrapped one
- value for the Content-Encoding HTTP header
- a function that writes to the output stream (this lets you control whether you write to a bufferedwriter or concatenate to a Stringbuilder)

It would look like this:

```clojure
(def gzip-profile
  ;; we use a string builder approach so we just add gzip and an OutputStreamWriter
  {:wrap (fn[^OutputStream os] (-> os ->gzip-os ->os-writer))
   :encoding "gzip"
   ;; the SDK provides a write function for using a StringBuilder to concatenate an SSE event's string
   :write! (->write-with-temp-buffer!)})

(def gzip-buffered-writer-profile
  ;; here we use a buffered writer so we add it to the wrapping function
  {:wrap (fn[^OutputStream os] (-> os ->gzip-os ->os-writer ->buffered-writer))
   :encoding "gzip"
   ;; The SDK also provides a function to write to a BufferedWriter
   :write! write-to-buffered-writer!})
```

With this you can control everything. You can replace the `->gzip-os` by a brotli one for instance.

This way the SDK would let you do almost whatever you want while providing you with helper functions to do it.
Also it would provide default "write-profiles" ready to use. The SDK wouldn't let you manage buffer size by using options, you either use the defaults or make your own output stream wrapper, setting these options however you want.

@Ramblurr @andersmurphy What do you think?

JeremS: I went ahead and implemented the `write profile` approach, I changed the docs to reflect this. The animation example can now use brotli.

There are some issues with it though. For one and done SSE events the compression works fine. For a stream of events the behavior can be weird. With ring jetty lots of events stay in a buffer until they are flushed all at once, even though the SDK flushes after writing each event. With http-kit brotli compression behaves a bit better but it's not working as well as with gzip. This is probably a skill issue on my part.

Still with the new approach, users can control the outputstream however they want.

Ramblurr:
@JeremS I just noticed the code snippets for copy/pasting under the Installation header in the readme have unbalanced {}s.

JeremS:
> @JeremS I just noticed the code snippets for copy/pasting under the Installation header in the readme have unbalanced {}s.

Fixed and I caught some typos too. Thank you for catching it! @Ramblurr

Ramblurr:
I've kicked the tires on the write-profile feature, and it is working as advertised.

Some thoughts (not specific to this PR but about the clojure SDK in general)

- Are we going to say it is up to the developer to parse `Accept-Encoding`  from requests and choose the appropriate content encoding?
  - I think the SDK should handle that (it already gets the req map) and then it can choose a write profile default -> big win for the developer as they don't have to think about compression (until they want to)
- Its a little strange that the `:on-open`/`:on-close` keywords are unqualified but the write-profile uses `:d*.sse/write-profile` It took me some time debugging why my naive `:write-profile ..` attempt didn't work.
- Documenting some "obvious" things (but weren't obvious to me):
  - The expected return type of on-open/on-close -> it doesn't matter, its not used?
  - The bodies of on-open/on-close, are they running in threads or in go blocks, can those bodies block doing I/O or crunch the cpu? 
  - Can on-open/on-close throw exceptions? What happens when they do?
- The SDK readme could use some polishing:
  - Document usage of write-profiles
  - Cleanup the usage of Notes/Important -> they make the document difficult to read (this is unfortunately a symptom of how Github renders them)
  - We should publish to clojars eventually, not just so it's straightforward to install, but also because we will get cljdoc.org for free, which would be very useful!

JeremS:
> I've kicked the tires on the write-profile feature, and it is working as advertised.

That's good!

> Some thoughts (not specific to this PR but about the clojure SDK in general)
> 
>     * Are we going to say it is up to the developer to parse `Accept-Encoding`  from requests and choose the appropriate content encoding?
>       
>       * I think the SDK should handle that (it already gets the req map) and then it can choose a write profile default -> big win for the developer as they don't have to think about compression (until they want to)

I think this a concern too far, the SDKs are meant to facilitate sending SSE events formatted the way datastar wants them.
Adding a way to compress the response is already a bit beyond the ADR we all follow. It led to the write profile functionality which gives a lot of freedom to a user without the need to re-implement the `SSEGenrator` protocol and a `->sse-response` function themselves. I believe it's a good addition.

In `SSE-design-notes.md` I discuss some considerations about SSE compression and buffering. When creating a write profile you may want to consider where you want to use buffers and of what size. You may want to add compression which means more buffers, context window config... You also need to consider how the ring adapter you are using works (Http-kit gives you an `AsyncChannel` not an `OutputStrem`). 

If we did content negotiation directly in the SDK, I feel we'd need to provide some solid API to handle all of these options cleanly. We also would need to provide some escape hatch.

I think this kind of functionality is great, but I believe it should come in it's own library outside of the SDK. Such a library could give you a wrapper function for the `->sse-response` function, a middleware basically. It would look at the `Accept-endocing` header. It would choose a profile based on that and other factors. Then would call the `->sse-response` function from the SDK with the selected profile. If such a library existed it would be mentioned in the SDK's README for sure.
 
>     * Its a little strange that the `:on-open`/`:on-close` keywords are unqualified but the write-profile uses `:d*.sse/write-profile` It took me some time debugging why my naive `:write-profile ..` attempt didn't work.

You are absolutely right, it is not consistent. Historicaly I went namespace qualified keywords for all the SSE option. Then when designing the `->sse-response` api I went `:on-open` and `:on-close` inspired by Http-kit. I didn't foresee adding other options at the time. 

I am of a mind to namespace qualify everything even though it is a breaking change. What to you think?

>     * Documenting some "obvious" things (but weren't obvious to me):
>       
>       * The expected return type of on-open/on-close -> it doesn't matter, its not used?
>       * The bodies of on-open/on-close, are they running in threads or in go blocks, can those bodies block doing I/O or crunch the cpu?
>       * Can on-open/on-close throw exceptions? What happens when they do?
> 
>     * The SDK readme could use some polishing:
>       
>       * Document usage of write-profiles
>       * Cleanup the usage of Notes/Important -> they make the document difficult to read (this is unfortunately a symptom of how Github renders them)

We need better docs. There is already an explanation of write profiles in `SSE-design-notes.md` I should extract it in its own documentation file.

>       * We should publish to clojars eventually, not just so it's straightforward to install, but also because we will get cljdoc.org for free, which would be very useful!

That is the plan, I refrained from working on this for 2 reasons:
- the SDK needs some polishing, and the artificial barrier to entry that is git deps wrongly lets me think I have some time before committing  to the API.
- I need to put together an automatic release process. Maybe by doing what other SDKs do, that is mirroring the SDK's code in its own repo and having some github task that will push automatically to a maven repository. I could use some help with this.

I think that after this PR is done, this is the next item to tackle. 

@Ramblurr Thanks for the feed-back!  I really appreciate your taking the time to review the code that thoroughly. Please let me know what you think!

JeremS:
I have remade the docs taking into account @Ramblurr remarks:

- the readme should be clearer
- write profile functionality is document in docstrings and in a file in the doc folder
- the behavior and expectations for each callback are documented in docstrings

I have also made a pass at all keywords to get consistent naming for all `->sse-response` options. The `:on-open` and `:on-close` callbacks are still acceptable but are marked deprecated. `:d*.sse/on-open` and `:d*.sse/on-close` are now the preferred way. This means no breaking change for now.

At this point I think this PR is ready. 

@andersmurphy @Ramblurr If you don't see any more changes that should be made I'd like to submit for integration in the main repo.

Cheers,

sdk/clojure/deps.edn
```diff
@@ -1,13 +1,25 @@
 {:paths ["sdk/src/main"]
  
- :deps {io.github.paintparty/fireworks {:mvn/version "0.10.4"}}
- 
+ :deps {io.github.paintparty/fireworks      {:mvn/version "0.10.4"}
+        com.taoensso/telemere               {:mvn/version "1.0.0-RC3"}
+        com.aayushatharva.brotli4j/brotli4j {:mvn/version "1.18.0"}}
```
andersmurphy, line 5: Are these now runtime dependencies for the SDK?

sdk/clojure/doc/SSE-design-notes.md
```diff
@@ -0,0 +1,117 @@
+# SSE, Buffering, Design considerations
```
andersmurphy, line 1: Really liked this doc

sdk/clojure/src/dev/examples/animation_gzip/brotli.clj
```diff
@@ -0,0 +1,48 @@
...
+    [starfederation.datastar.clojure.adapter.common :as ac])
+  (:import
+    com.aayushatharva.brotli4j.Brotli4jLoader
+    [com.aayushatharva.brotli4j.encoder  Encoder$Parameters
+      Encoder$Mode BrotliOutputStream]
+    [java.io OutputStream]))
+
+;; This code is adapted from https://github.com/andersmurphy/hyperlith/blob/master/src/hyperlith/impl/brotli.clj
+;; Thank you Anders!
+
```
andersmurphy, line 12: ❤️

sdk/clojure/src/test/core-sdk/starfederation/datastar/clojure/adapter/common_test.clj
```diff
@@ -0,0 +1,160 @@
+(ns starfederation.datastar.clojure.adapter.common-test
+  (:require
+    [starfederation.datastar.clojure.api :as d*]
+    [starfederation.datastar.clojure.adapter.common :as ac]
+    [starfederation.datastar.clojure.adapter.test :as at]
+    [lazytest.core :as lt :refer [defdescribe describe specify expect]])
```
andersmurphy, line 6: How are you finding lazytest?

andersmurphy (approved):
A few minor comments, but looks like a solid improvement. Would still be good to get @Ramblurr take as it's a big PR, so the more eyes the better!

sdk/clojure/deps.edn
JeremS: Yes the main deps.edn covers dependencies I use while working on the SDK and running tests. The SDK proper has 0 deps, the adapter specific libs and the malli schema only depend on the bare minimum.

sdk/clojure/src/test/core-sdk/starfederation/datastar/clojure/adapter/common_test.clj
JeremS: I like it. I have used the metatada feature to filter tests when running test continoulsy. The way fixtures are implemented works well. It's a good tool and there is effort in the community to support it.

Ramblurr (approved):
:+1:  The docs are looking really good now. Being consistent with the kws is a welcome change. 

My only suggestion is that we remove the deprecated keywords before any sort of 1.0 release.

JeremS:
> My only suggestion is that we remove the deprecated keywords before any sort of 1.0 release.

That is the plan!

@bencroker I think we can go ahead now!

bencroker:
Merged, great work, thanks!

JeremS:
@bencroker Thank you!

PR #748, opened 2025-03-08
Add exports for bundles and plugins

mat02:
This PR adds exports for plugins, bundles and types, so npm package can be used more flexibly.
Closes  ##742.

Example of usage:
```
import { load } from '@starfederation/datastar/bundles/datastar-core' // import only core bundle
import { Clipboard } from '@starfederation/datastar/plugins' // import arbitrary official plugins
import { type ActionPlugin, PluginType }  from '@starfederation/datastar/types' // import types to add own plugins

const customPlugin:  ActionPlugin = {
    type: PluginType.Action,
    name: 'testMe',
    fn: () => {
        console.log('Hello from custom action!')
    },
}

load(Clipboard, customPlugin)
```

### DIFF
--- a/library/package.json
+++ b/library/package.json
@@ -18,6 +18,24 @@
   ],
   "module": "dist/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    },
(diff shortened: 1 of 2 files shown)


delaneyj (approved):

bencroker:
Merged, thanks!

PR #756, opened 2025-03-11
Go SDK WithCompression option for NewSSE

Yacobolo:
Wanted compression for my SSE. So added it to the SDK.
Also wanted to add support for brotli - currently i made it using gzip. I have tested it and it works however - i think i need to safely close the gzip writer, just could not figure out where.


bencroker:
Please switch to a PR draft until this is ready for review.

sdk/go/sse.go
```diff
@@ -10,18 +10,69 @@ import (
...
+
+func WithZstd() SSEOption {
+	return func(sse *ServerSentEventGenerator) {
+		comp, _ := zstd.New()
+		sse.w = comp.Get(sse.w)
+		sse.encoding = zstd.Encoding
+	}
+}
+
+func WithDefaultGzip() SSEOption {
```
delaneyj: Closer!  But these should take options as well... you should be able to config, not just support the default levels

delaneyj (changes requested):

Yacobolo:
@delaneyj  updated the code. Let me know if it was this that you meant 👍

delaneyj (approved):

bencroker:
Merged, thanks!

Issue #769, opened 2025-03-16
Break out the Python SDK's `responses` module into a package


jmoppel:
### Feature Request

The current implementation of the responses is in a single module. This leads to somewhat awkward code like:

```python
from datastar_py import consts
from datastar_py.responses import DatastarDjangoResponse  # could be improved

class FooDetailView(DetailView):
    model = Foo
    def get(self, request, *args, **kwargs):
        datastar_request = request.headers.get("datastar-request")
        if not datastar_request:
            return super().get(request, *args, **kwargs)
        
        # datastar specific code happens...

        return DatastarDjangoResponse(generate_fragments)  # could be improved
```

In particular, the last line mentioning the name of the framework (`Django`) feels less than ideal. I imagine it would feel equally janky in other frameworks. I think an improvement we could do something like:

```python
from datastar_py import consts
from datastar_py.responses.django import DatastarHttpStreamingResponse  # changed

class FooDetailView(SoftDeleteSingleMixin, DetailView):
    model = Foo
    def get(self, request, *args, **kwargs):
        datastar_request = request.headers.get("datastar-request")
        if not datastar_request:
            return super().get(request, *args, **kwargs)
        
        # datastar specific code happens...

        return DatastarHttpStreamingResponse(generate_fragments)  # changed
```

`HttpStreamingResponse` is a part of Django, therefore `DatastarHttpStreamingResponse` feels very idiomatic. Breaking the responses into their own modules would allow similar adjustments to make `datastar_py` appear more idiomatic to each framework.

This would result in quite  few very small modules, but maybe it is worth the change as it also allows additional, framework specific additions going forward.

I'm happy to submit a PR to break this `responses` into a package, but wanted to check in to see if others are open to this change or perhaps some other better idea.

Paging @lllama 

### Datastar Version

v1.0.0-beta.9

bencroker:
Paging @lllama.

gazpachoking:
Agreed on this. I think it would also be nice if the classes had the same signature as the native streaming response class of the framework (as in #766)

gazpachoking:
Actually, do we even need responses to be a package? What about just having modules for each framework? i.e. `from datastar_py.django import DatastarStreamingHttpResponse`

anastygnome:
> Actually, do we even need responses to be a package? What about just having modules for each framework? i.e. `from datastar_py.django import DatastarStreamingHttpResponse`

[...]

PR #770, opened 2025-03-16
Split various python SDK frameworks to different modules

gazpachoking:
Marking this as a draft since I'm looking for feedback about this approach. This hasn't been fully tested yet, so probably isn't suitable to merge as-is. I'm happy to do some more work if people like this approach, or happy for anyone to run with their own solution if inspired by these ideas.

This is an example of the idea from #769, mixed with some of my ideas from #766.

Main points:
- Split the different frameworks up into different files.
- Names the responses the same as they are in their parent framework with the addition of 'datastar'
- Now keeps the same signature for responses as they are in their parent framework. (this also fixes the problem pointed out in #766, where you could not pass arguments to your sse generator functions)
- For frameworks where we are providing a response class, mix in the classmethods from `ServerSentEventGenerator` for more ergonomic usage.
- Added starlette framework. Both fastapi and fasthtml use this under the hood, so they can use that response directly.

Possible further changes:
- [x] name the sanic helper `datastar_respond` to closer mimic native? (native is `response = request.respond()` proposed change would be `response = datastar_respond(request)`


gazpachoking:
@lllama I hear you are the python SDK guy, interested in your thoughts here.

gazpachoking:
Okay, I tested all the examples work, except django. I couldn't get that one to work. It didn't like the keep-alive header, or an async generator for the response, neither of which have anything to do with these changes, so I left it for someone who knows something about django.

bencroker:
Review please, @lllama.

jmoppel:
> Okay, I tested all the examples work, except django. I couldn't get that one to work.  It didn't like the keep-alive header, or an async generator for the response

The problem you experienced getting Django running should be fixed by #787

lllama:
Looks good overall. I'm 50/50 on moving the responses to their own submodules but it could be useful in the future if extra functionality for specific frameworks is added.

LGTM :shipit:

bencroker (approved):

bencroker:
Approved, please resolve the merge conflict.

bencroker:
Merged, thanks!

PR #773, opened 2025-03-17
Make consts in python SDK more idiomatic. fix #767

### LINKED ISSUE #767: Python SDK `consts` violates PEP formatting guidlines
### Bug Report

There are various formatting and code quality issues with the code generated in the `consts` module (Python SDK). There may be other similar issues, but I only looked at consts for now since it is generated code.

### Variables

PEP-8 specifies that variable constants [should be in all caps with an underscore separating words](https://peps.python.org/pep-0008/#constants). Some variables follow this convention (e.g., `DATASTAR_KEY` and `VERSION`), but most others don't are (incorrectly) in `CapWords` (i.e., PascalCase). A few random examples include:

&cross; Current Code
```python
DefaultFragmentsSettleDuration = 300
DefaultExecuteScriptAttributes = "type module"
DefaultFragmentsUseViewTransitions = False
```

&check; Corrected Variable Names
```python
DEFAULT_FRAGMENTS_SETTLE_DURATION = 300
DEFAULT_EXECUTE_SCRIPT_ATTRIBUTES = "type module"
DEFAULT_FRAGMENTS_USE_VIEW_TRANSITIONS = False
```

### Enums

Because they represent constants, Enum member names (i.e., the variables inside an enum class) [should also be all caps with underscores](https://docs.python.org/3/howto/enum.html) per PEP-8 (see constants above).

Additionally, the Enum member's names, pointlessly duplicate the class name of the ENUM which hurts readability, and is a violation of [PEP-0020](https://peps.python.org/pep-0020/) "Readability counts". The name for each enum member should probably be the same as their value, and capitalized like a constant (upper case w/words separated by an underscore). I grant that this is a bit subjective, but nonetheless decided to mention it because some code reads rather poorly (IMO).

&cross; Current Code
```python
class FragmentMergeMode(StrEnum):
    # Morphs the fragment into the existing element using idiomorph.
    FragmentMergeModeMorph = "morph"  # bad caps and name could be better

    # Replaces the inner HTML of the existing element.
    FragmentMergeModeInner = "inner"

    # Replaces the outer HTML of the existing element.
    FragmentMergeModeOuter = "outer"

    # Prepends the fragment to the existing element.
    FragmentMergeModePrepend = "prepend"

    # Appends the fragment to the existing element.
    FragmentMergeModeAppend = "append"

    # Inserts the fragment before the existing element.
    FragmentMergeModeBefore = "before"

    # Inserts the fragment after the existing element.
    FragmentMergeModeAfter = "after"

    # Upserts the attributes of the existing element.
    FragmentMergeModeUpsertAttributes = "upsertAttributes"


# instance example
morph = FragmentMergeMode.FragmentMergeModeMorph

```

&check; Corrected and Improved Enum Members
```python
class FragmentMergeMode(StrEnum):
    # Morphs the fragment into the existing element using idiomorph.
    MORPH = "morph"

    # Replaces the inner HTML of the existing element.
    INNER = "inner"

    # Replaces the outer HTML of the existing element.
    OUTER = "outer"

    # Prepends the fragment to the existing element.
    PREPEND = "prepend"

    # Appends the fragment to the existing element.
    APPEND = "append"

    # Inserts the fragment before the existing element.
    BEFORE = "before"

    # Inserts the fragment after the existing element.
    AFTER = "after"

    # Upserts the attributes of the existing element.
    UPSERT_ATTRIBUTES = "upsertAttributes"


# instance example
morph = FragmentMergeMode.MORPH
```

### Miscellaneous formatting issues.

There are several minor PEP-8 formatting issues. These could the fixed by introducing Python linters like Black and Flake-8. I'm happy to introduce a PR to add `black` and `flake8` to the library's `pyproject.toml` and perhaps CI if that's welcome. Below are a few examples of the small formatting changes black/flake8 would fix.

PEP-8 [pet peeves violation](https://peps.python.org/pep-0008/#pet-peeves)
&cross; Current Code
```python
VERSION                   = "1.0.0-beta.9"  # extraneous spaces
```

&check; Removed extra spaces.
```python
VERSION = "1.0.0-beta.9"  # extraneous spaces
```

PEP-8 [block comment violation](https://peps.python.org/pep-0008/#block-comments) 
&cross; Current Code
```python
#endregion EventType
```

&check; Added space at beginning of comment
```python
# endregion EventType
```

### Datastar Version

v1.0.0-beta.9

gazpachoking:
Addresses the consts in the python SDK to follow traditional python conventions, as detailed in #767


bencroker:
Please review, @lllama.

gazpachoking:
Also made sure consts output matches ruff/black formatter (except there is no newline at the end of the file, seems the build process strips that no matter what)

lllama:
LGTM :shipit: 

Very nice @gazpachoking

sdk/python/src/datastar_py/consts.py
```diff
@@ -1,96 +1,102 @@
...
-#region The mode in which a fragment is merged into the DOM.
+
+# region Enums
+
+
+# region The mode in which a fragment is merged into the DOM.
 class FragmentMergeMode(StrEnum):
     # Morphs the fragment into the existing element using idiomorph.
-    FragmentMergeModeMorph = "morph"
+    MORPH = "morph"
```
bencroker, line 11: Please revert to the convention used by all SDKs, in this case it should be `FRAGMENT_MERGE_MODE_MORPH`.

sdk/python/src/datastar_py/sse.py
```diff
@@ -139,25 +139,25 @@ def execute_script(
...
 
         data_lines.extend(
-            f"data: {consts.ScriptDatalineLiteral} {script_line}"
+            f"data: {consts.SCRIPT_DATALINE_LITERAL} {script_line}"
             for script_line in script.splitlines()
         )
 
         return ServerSentEventGenerator._send(
-            consts.EventType.EventTypeExecuteScript,
+            consts.EventType.EXECUTE_SCRIPT,
```
bencroker, line 160: Please revert to the convention used by all SDKs, in this case it should be `EVENT_TYPE_EXECUTE_SCRIPT`.

sdk/python/src/datastar_py/consts.py
gazpachoking: Even though it's used with the namespace already, i.e. `FragmentMergeMode.MORPH`? Having it doubled feels redundant and very long. `FragmentMergeMode.FRAGMENT_MERGE_MODE_MORPH`

gazpachoking: I had a peek at a few other SDKs and it seems they follow this convention.
https://github.com/starfederation/datastar/blob/db376dfbe4d100ea997c932f1301fbe72e08dfe8/sdk/java/core/src/main/java/starfederation/datastar/enums/FragmentMergeMode.java#L6-L9
https://github.com/starfederation/datastar/blob/db376dfbe4d100ea997c932f1301fbe72e08dfe8/sdk/dotnet/src/Consts.fs#L66-L69

bencroker: Yeah you’re right, better to avoid duplication.

sdk/python/src/datastar_py/sse.py
bencroker: Ignore me.

bencroker (approved):

bencroker:
Merged, thanks!

gazpachoking:
Thanks!

PR #775, opened 2025-03-17
Home grown signals library

delaneyj:
With much back and forth with @jmstevers here is an initial implementation that passes all tests and is between 70% and 6113% faster.  This is mostly from simplifying how dependencies/subscriptions are checked and how often.


library/src/engine/engine.ts
```diff
@@ -340,6 +343,13 @@ function genRX(
   if (signalNames.length) {
     // Match any valid `$signalName` followed by a non-word character or end of string
     const signalsRe = new RegExp(`\\$(${signalNames.join('|')})(\\W|$)`, 'gm')
+    for (const match of userExpression.matchAll(signalsRe)) {
+      const signalName = match[1]
+      const signal = ctx.signals.signal(signalName)
+      if (signal) {
+        deps.push(signal)
+      }
```
bencroker, line 351: We should be able to perform the replace here too, right?

```ts
      if (signal) {
        deps.push(signal)
        userExpression = userExpression.replaceAll(
          `$${signalName}`,
          `ctx.signals.signal('${signalName}').value`,
        )
      }
```

bencroker:
Incredible work!! Should we consider extracting the signals lib into its own file? That could better set us up to maintain 🚀 in a separate repo.

library/src/plugins/official/logic/actions/setAll.ts
```diff
@@ -10,7 +11,7 @@ export const SetAll: ActionPlugin = {
   fn: ({ signals }, prefix: string, newValue) => {
     signals.walk((path, signal) => {
       if (!path.startsWith(prefix)) return
-      signal.value = newValue
+      ;(signal as Signal<any>).value = newValue
```
bencroker, line 14: Redundant semicolon?

library/src/plugins/official/logic/actions/toggleAll.ts
```diff
@@ -10,7 +11,7 @@ export const ToggleAll: ActionPlugin = {
   fn: ({ signals }, prefix: string) => {
     signals.walk((path, signal) => {
       if (!path.startsWith(prefix)) return
-      signal.value = !signal.value
+      ;(signal as Signal<any>).value = !signal.value
```
bencroker, line 14: Redundant semicolon?

library/src/plugins/official/logic/actions/setAll.ts
delaneyj: Oh my sweet summer child. JS wat my dear

bencroker: 🖕JS. Could we rewrite to the following for clarity?
```js
if (!path.startsWith(prefix)) {
  return
}
```

bencroker: Done.

PR #776, opened 2025-03-18
Improve the datastar expression guide

Ramblurr:
After chatting with Anders, I got a list of "things to know" about D* expressions.

This commit expands upon the current docs, adding more facts, examples and some structure.

### DIFF
--- a/site/static/md/guide/datastar_expressions.md
+++ b/site/static/md/guide/datastar_expressions.md
@@ -2,13 +2,11 @@
 
 Datastar expressions are strings that are evaluated by Datastar attributes and actions. While they are very similar to JavaScript, there are some important differences that make them more powerful for declarative hypermedia applications.
 
-- [Datastar Expressions](#datastar-expressions)
-  - [Basic Usage](#basic-usage)
-  - [Namespaced Signals](#namespaced-signals)
-  - [Multiple Statements and Formatting](#multiple-statements-and-formatting)
-  - [Event Context](#event-context)
-  - [Summary](#expression-summary)
-
+- [Basic Usage](#basic-usage)
+- [Namespaced Signals](#namespaced-signals)
+- [Multiple Statements and Formatting](#multiple-statements-and-formatting)
+- [Event Context](#event-context)
+- [Summary](#expression-summary)
 
 ## Basic Usage
 
--- a/site/static/md/guide/datastar_expressions.md
+++ b/site/static/md/guide/datastar_expressions.md
@@ -45,7 +45,22 @@ Unlike JavaScript variables, signal names, when kebab cased, can dashes that wou
 
 This works because behind the scenes, Datastar is transforming the datastar expression into the javascript `ctx.signals.signal('user-name').value = "Alice"` - the signal name is passed as a string, not used as a direct JavaScript identifier.
 
-This should help clarify what Datastar is doing behind the scenes, how signals are evaluated in expressions, and also why expressions are limited in scope.
+Normal JavaScript operators are of course available in Datastar expressions. This includes (but isn't limited to) `&&`, `||`, and the ternary operator. These last three can be useful when reacting to signal changes.
+
+For example, the following would only trigger a post action if the signal is logically true.
+
+```html
+<button data-on-click="$landingGearRetracted && @post('/launch')"></div
+
+<!-- or if the signal has a state value --->
+<button data-on-click="$landingGearState == 'retracted' && @post('/launch')"></div
+```
+
+The ternary operator is useful to choose among two options:
+```html
+<div data-attr-class="$theme == 'dark' ? 'bg-black text-white' : 'bg-white text-black'"
+```
+
 
 ## Namespaced Signals
 
--- a/site/static/md/guide/datastar_expressions.md
+++ b/site/static/md/guide/datastar_expressions.md
@@ -1,6 +1,16 @@
 # Datastar Expressions
 
-Datastar expressions are strings that are evaluated by Datastar attributes and actions. 
+Datastar expressions are strings that are evaluated by Datastar attributes and actions. While they are very similar to JavaScript, there are some important differences that make them more powerful for declarative hypermedia applications.
+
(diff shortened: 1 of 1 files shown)


Ramblurr:
@andersmurphy since this was based off info from you, maybe you could look? 

Screenshot 2025-03-18 at 08-04-27 Datastar Expressions

site/static/md/guide/datastar_expressions.md
```diff
@@ -26,27 +36,42 @@ return (()=> {
 })()
 ```
 
-This should help clarify what Datastar is doing behind the scenes, how signals are evaluated in expressions, and also why expressions are limited in scope. 
+Unlike JavaScript variables, signal names, when kebab cased, can dashes that would be invalid in standard JavaScript identifiers:
```
bencroker: I’d rather not document a convention we don’t want to promote – camel-casing signal names is the better convention. Can you please remove this paragraph and example?

bencroker:
Thanks for the suggestions! I’d really like to see the convention of using `&&` explained, for example:

```html
<button data-on-click="$foo == 1 && @get('...')"></div
```

See also:
https://data-star.dev/examples/bind_keys
https://data-star.dev/examples/delete_row
https://data-star.dev/examples/dialogs_browser

Ramblurr:
Ah that is definitely an oversight :) I'll add it.

Ramblurr:
> Thanks for the suggestions! I’d really like to see the convention of using `&&` explained, for example:
> 
> ```
> <button data-on-click="$foo == 1 && @get('...')"></div
> ```
> 
> See also: https://data-star.dev/examples/bind_keys https://data-star.dev/examples/delete_row https://data-star.dev/examples/dialogs_browser

Added @bencroker let me know if that is sufficient.

bencroker:
Thanks! Please also see my review comment above.

Ramblurr:
Sorry, I don't see any review comments in this PR. The only comment I see that is actionable is your one asking to document "&&", which I have.

Could you please clarify?

site/static/md/guide/datastar_expressions.md
bencroker: Resolved.

bencroker (approved):

bencroker:
I opened a review, not sure why it doesn’t appear for you above. Never mind though, I made some tweaks and merged. Great work, thanks!

Ramblurr:
@bencroker I hope I'm not being a pest:

image


..but that bit you removed explaining how kebabed case signal names can be used in datastar expressions was my original intent for even opening this PR. For someone who knows javascript and knows that identifiers cannot have `-` in them, it is not at all obvious that you reference kebab cased signals like I depicted.

If you didn't like how long that section was, I'm fine with shortening it, but the documentation should say tha *somewhere*.

Furthermore, the bit about how datastar transforms the signal name into `ctx.signals.signal...` was originally part of that document, and I think very relevant for users to understand what d* is doing, it isn't magic.

Can we re-add some bits of that?

bencroker:
Part of my review comment was that I’d rather not document a convention we don’t want to promote – camel-casing signal names is the better convention. Just because something is possible, it doesn’t mean it must be documented.

Transforming the signal name into `ctx.signals.signal...` was and is already documented, so I see no reason to reiterate it.

screenshot-ycbQ81vg@2x

Ramblurr:
> Part of my review comment was that I’d rather not document a convention we don’t want to promote – camel-casing signal names is the better convention. Just because something is possible, it doesn’t mean it must be documented.

The camelCase hegemony strikes again! 

Please keep in mind that camel casing might be a better convention for you in PHP or Go. It is not the convention in Clojure, where we use kebab case. 

You've said on YT something along the lines of  datastar being a backend-focused frontend framework. Clojure folks will want to continue using their kebab cased convention, and I hope d* doesn't make that more difficult than necessary in the name of convention.

But anyways, us kebab-caser being in the minority, I understand your point. Maybe this something we can add to the clojure sdk docs, because every clojurist will struggle with casing in their minutes of using datastar.

> Transforming the signal name into `ctx.signals.signal...` was and is already documented, so I see no reason to reiterate it.

Ah yes, I missed that one. Agreed.

bencroker:
> Please keep in mind that camel casing might be a better convention for you in PHP or Go. It is not the convention in Clojure, where we use kebab case.

It’s a better convention in Datastar (on the frontend), simply due to it being the default when using `data-signals-foo-bar`. Can you explain why it would make any difference on the backend?

Ramblurr:
One of the beautiful things about datastar is that while programming there is no frontend, the code writing is all happening "on the backend".

In clojure we generally do not have separate template html files. Our view code is just clojure code, which is full of kebab cased variable names.

I fetch a "user" from the database and it looks like this:
```clojure
{ :user-name "Ramblurr"
  :email "foo@example.com"
  :first-name "Casey"}
```

If I want to render a form so the user can edit their name, I write a function that looks like this:

```clojure
(defn user-edit-view [user-map] ;; user-map is a map of the user data I showed above
  [:div {:data-signals (json/serialize user-map)}
   [:input {:type :text :id :user-name :name "user-name" :data-bind "user-name"}]
   [:input {:type :text :id :first-name :name "first-name" :data-bind "first-name"}]
   [:button {:data-on-click "@post('/user/edit')"} "Save"]])
```

The serialize function produces JSON which data star happily accepts:

```json
{
  "user-name": "Ramblurr",
  "email": "foo@example.com",
  "first-name": "Casey"
}
```

Then here is the POST handler:
```clojure
(defn user-edit-handler [request]
  (let [submitted-signals (get-posted-params request) ;; <--- this is will be the map of signals,
        user              (get-user-from-database)]   ;;      in kebab case, just like my backend expects.
    (update-user user submitted-signals)
    {:status 204}))

```

Datatar posts back json that looks similar to the above, but with the user's edits. 
No where did I have to transform `user-name` to `userName` or `first-name` to `firstName`  on the way out, nor the reverse when handling the POST.

bencroker:
Thanks for clarifying! Please create a new PR for it and I’ll merge (sorry about the double PR).

PR #777, opened 2025-03-18
Make python sdk compatible with python <3.11

gazpachoking:
`StrEnum` was introduced in python 3.11. In order to make the sdk compatible with all still supported python versions (>=3.9). We can recreate that class ourself.


bencroker:
Thanks for the suggestion. Python 3.11 was released in October 2022, so I see no reason to backport functions used by the SDK.

bencroker:
Reopened per request. Please review, @lllama.

lllama:
LGTM. :shipit:

bencroker (approved):

bencroker:
Merged, thanks!

PR #778, opened 2025-03-19
Clarify that data-computed is a pure function

Ramblurr:
Also make explicit that `data-on-signals-change` is the alternative for
side effects.

### DIFF
--- a/site/static/md/reference/attribute_plugins.md
+++ b/site/static/md/reference/attribute_plugins.md
@@ -112,19 +112,19 @@ Computed signals are useful for memoizing expressions containing other signals. 
 
 `data-computed` is a pure reactive function, this has several implications:
 
-1. If a computed signal is not consumed, then the computation will not execute
-2. Computed signals must not be used for side-effects (i.e., changing other signals, actions, JavaScript functions, etc)
+1. If a computed signal is not consumed, then the computation will not execute.
+2. Computed signals must not be used for performing actions (changing other signals, actions, JavaScript functions, etc.).
 
 ```html
 <!-- This computation will never execute because $foo is not used anywhere -->
 <div data-computed-foo="$bar + $baz"></div> <!-- WRONG -->
 
-<!-- computed signals must *not* be used for side effects -->
+<!-- Computed signals must *not* be used for side effects -->
 <div data-computed-qux="@post('/qux'); 'quxed'"></div> <!-- WRONG -->
 <div data-computed-foo="$bar++"></div> <!-- WRONG -->
 ```
 
-If you find yourself wanting to side-effect in reaction to a signal change, refer to the [`data-on-signals-change`](#special-events) special event in the [`data-on` plugin](#data-on)
+If you find yourself wanting to perform some action in reaction to a signal change, refer to the [`data-on-signals-change`](#special-events) attribute in the [`data-on` plugin](#data-on)
 
 #### Modifiers
 
--- a/site/static/md/reference/attribute_plugins.md
+++ b/site/static/md/reference/attribute_plugins.md
@@ -110,6 +110,22 @@ Computed signals are useful for memoizing expressions containing other signals. 
 <div data-text="$foo"></div>
 ```
 
+`data-computed` is a pure reactive function, this has several implications:
+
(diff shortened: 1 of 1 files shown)


bencroker:
Thanks for the suggestions. I’m not sure how beneficial the clarification around `data-computed` is, however. Nobody has expressed confusion around the clarifications you make. You also use language like “pure reactive function” and “side-effects” which are not incorrect, but might be unfamiliar to people not approaching Datastar from SPA frameworks.

Ramblurr:
> Nobody has expressed confusion around the clarifications you make.

It is early days for datastar, there has to be a first time :) This plugin has only existed a few months.

> You also use language like “pure reactive function” and “side-effects” which are not incorrect, but might be unfamiliar to people not approaching Datastar from SPA frameworks.

This language is used by @delaneyj not only in our [discord conversation](https://discord.com/channels/1296224603642925098/1333801016642244739/1351641628997324844) about it, but also on podcasts and interviews where he is spreading the good word.

And for the record, I am not approaching Datastar from an SPA framework, I've avoided them and plan to continue.  

Datastar is (in part) a signals library. Avoiding using terms like "reactive" and "side-effects" in the *reference* documentation seems rather silly? If this was a How-to or beginner's introduction, then sure I would avoid dropping terms like that without surrounding explanation. Do you have a suggestion for how to re-word it? I'm happy to tweak it!

(And in my opinion "side effects" is a phrase any programmer should know).

Further reasons why I think this documentation change is warranted: `data-computed` exists as its own top-level plugin. It is the 2nd plugin in the documentation. It allows you to execute code when something changes. 

Often when "something changes" you might want to execute some other code that "touches the world" (aka a side-effect), data-computed is NOT the plugin you should reach for.

Which plugin should you reach for? Well what you really want is documented, not as a top-level plugin, but as a "special" version of another plugin `data-on`. Which does not get a top-level full treatment, instead it gets one bullet point out of 4.

site/static/md/reference/attribute_plugins.md
```diff
@@ -110,6 +110,22 @@ Computed signals are useful for memoizing expressions containing other signals.
 <div data-text="$foo"></div>
 ```
 
+`data-computed` is a pure reactive function, this has several implications:
+
+1. If a computed signal is not consumed, then the computation will not execute
```
bencroker: > If a computed signal is not consumed, then the computation will not execute

Are you sure about this? It doesn’t sound accurate to me.

Ramblurr: Yes it is accurate. Because computed signals are pure functions, they should not have side effects (otherwise they aren't pure!), which means they can be lazily computed.

This is how I originally discovered that `data-computed` is not to be used for side effects. I wanted to do a side-effect in a computed signal that otherwise wasn't used. See the  [discord conversation](https://discord.com/channels/1296224603642925098/1333801016642244739/1351641628997324844) where @delaneyj  very pointedly makes me aware of that fact.

Anyways you can try it out for yourself, define a `data-computed-my-computation="console.log('foobar') && 'woot'"` Then if you do not use the `$myComputation` signal, the log will not fire. but if you do `data-text="$myComputation"` you'll see the text is `woot` and a console log fires.

bencroker: I tested it and see `foobar` in the console, as expected. Please retest on your end.

Ramblurr: Please see this example:


```html
<div data-signals-woot="'woot'"
     data-computed-my-computation="console.log('computed'); $woot">
    <span data-text="$myComputation"></span>
</div>
```

versus

```html
<div data-signals-woot="'woot'"
     data-computed-my-computation="console.log('computed'); $woot">
</div>
```

The first snippet will print "computed" to the log, and the span will show "woot". The second example will not have anything printed to the log, because the computation is not run because `$myComputation` is not used.

bencroker: Yes you’re right. Unaware, I was testing using a template that contained `ctx.signals.JSON()`.

bencroker:
All good points! I’ve made some tweaks, changed “side-effects“ to the more generic “perform actions”, and merged.

Thank you once again!

PR #779, opened 2025-03-19
Improve casing and modifier docs

Ramblurr:
- Expand the attribute casing documentation
    
    The previous top-line message "data-* attributes are case-insensitive", while
    true, was very confusing for new users because it is not immedietly clear that
    sentence is referring to the HTML spec and not the library of the docs they are reading.
    
    So I've clarified the top-line message, and then expanded the explanation in a
    way I hope makes it easier to understand the first time.
- Explain how to use modifiers with `data-signals` in the object notation variant

### DIFF
--- a/site/static/md/reference/attribute_plugins.md
+++ b/site/static/md/reference/attribute_plugins.md
@@ -11,7 +11,9 @@ Datastar provides the following [`data-*`](https://developer.mozilla.org/en-US/d
 
 ### Attribute Order
 
-Note that `data-*` attributes are evaluated in the order they appear in the DOM. Elements are evaluated by walking the DOM in a depth-first manner, and attributes are processed in the order they appear in the element. This means that if you use a signal in a [Datastar expression](/guide/datastar_expressions), it must be defined _before_ it is used.
+<em>`data-*` attributes are evaluated in the order they appear in the DOM</em>.
+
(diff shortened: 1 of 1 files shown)


site/static/md/reference/attribute_plugins.md
```diff
@@ -95,6 +106,25 @@ Modifiers allow you to modify behavior when merging signals.
 ></div>
 ```
 
+When supplying signals in bulk with object notation, the modifiers still apply to the attribute name:
```
bencroker: What is the purpose of this? It appears to only show the `__ifmissing` modifier, rather than anything related to casing.

Ramblurr: This section of the docs explains 4 concepts:

1. data-signals-foo
2. data-signals=object notation format
3. case modifiers
4. ifmissing modifier

image

(screenshot from current `develop`, **not** my branch)

---

My addition starting with "When supplying signals.." is directly related to 1, 2, and 4. I don't see why it would have to be only about 3?

bencroker: Fair point!

bencroker: Fair point!

bencroker:
Made some minor tweaks and merged. Great work, thanks!

PR #787, opened 2025-03-22
Fix WSGI/ASGI issues in Django Example Project

jmoppel:
- Fixes the `keep-alive` `connection` header error when using Django's runserver (i.e., WSGI apps)
- Add views/urls to support a WSGI-only example.
- Makes ASGI the default and install/configure `daphne`, so we have an ASGI `runserver`
- Added a `pyproject.toml` to make it easier to install dependencies for the example
- Added a README for the example project mostly to help folks get WSGI working.

paging @lllama

### DIFF
--- a/examples/python/django/README.md
+++ b/examples/python/django/README.md
@@ -1,0 +1,26 @@
+# Datastar Django
+
+ASGI and WSGI examples of Datastar and Django in action.
+
+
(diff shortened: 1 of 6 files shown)


lllama:
LGTM

bencroker (approved):

bencroker:
Merged, thanks!

PR #800, opened 2025-03-28
fix(docs): Clarify file input type derived signals

dpc:
For a newcomer the fact that these signals are *set* is a bit magical, while the fact that they are sent seems kind of natural/obvious.


bencroker (approved):

bencroker:
Merged, thanks!

PR #808, opened 2025-03-31
Python SDK: Stop sending Connection header

gazpachoking:
This one probably needs some discussion. The goal here is to have the right headers for multiple HTTP versions. I was in the middle of another PR, which would change the API to pass in the request to all our response classes/functions so that we could detect the HTTP version in use. This would complicate and be a breaking change for most of the python frameworks in the SDK, as well as make the api for streaming responses mismatch with the frameworks we are integrating with. As I was reading the protocol guide/spec, it seems to me that the only version of http where we need the Connection header is 1.0, which I am tempted to believe we don't really need to support anymore. The spec says:
- http 2, Connection header prohibited
- http 1.1, Connection header defaults to keep-alive, which is what we want

Am I missing anything here, is http 1.0 on either the browser or python backend really a thing that is still in use? Is there some other reason why it's important to send this header with http 1.1 even though it's the default?

### DIFF
--- a/sdk/python/src/datastar_py/sse.py
+++ b/sdk/python/src/datastar_py/sse.py
@@ -6,7 +6,6 @@ import datastar_py.consts as consts
 
 SSE_HEADERS = {
     "Cache-Control": "no-cache",
-    "Connection": "keep-alive",
     "Content-Type": "text/event-stream",
 }
 


bencroker:
Any update on this, @gazpachoking ?

gazpachoking:
I think this change is good. The HTTP 1.1 spec says that persistent connections are the default, and only defines that 'Connection: close' should be sent to close them immediately. The only mention of sending 'Connection: keep-alive' is for compatibility with HTTP 1.0 clients, which I don't think we really need to worry about. In HTTP 2 and 3, the Connection header is forbidden, so this change makes us compatible with those.

gazpachoking:
Well, more discussion in discord landed on we might want this still to be compatible with some proxies. We'll need to change the python sdk api in order to become compatible with http 2/3 then.

bencroker (approved):

bencroker:
I agree that this change makes sense. We can discuss further measures in Discord.

Merged, thanks!

PR #809, opened 2025-03-31
Python SDK: Allow implicit merge-fragments/signals events and decorator for iterator responses

gazpachoking:
I've separated this PR into two separate ideas for discussion and consideration:
1. Allow implicit merge_fragments and merge_signals responses. This would allow you to return a string with an html fragment without wrapping it in `ServerSentEventGenerator.merge_fragment` Similarly a dictionary would implicitly be a merge_signals.
2. Add a decorator that wraps a generator function in the appropriate DatastarResponse class. This would mean your route handler can be the generator function directly, rather than having to have it separate from (or contained within) the route handling function.

Putting this up as a draft for discussion before more work. Some questions to answer:
- Do we want either of these things? I think it makes the common case of an endpoint that just merges fragments and signals super clean, but it's not that hard to do it normally, so maybe more explicit is better.
- Naming of the decorator. `sse_generator`?

TODO if we like this:
- [ ] Test django decorator
- [x] ~~Do a fasthtml decorator. The starlette one would work, but it needs the special bit to turn the result from fasthtml stuff into html.~~
- [ ] Update more examples? Not sure the best way to do examples, because I think we still want to show the regular method, as well as the convenience shortcuts.

### DIFF
--- a/examples/python/fastapi/app.py
+++ b/examples/python/fastapi/app.py
@@ -1,10 +1,12 @@
 import asyncio
 from datetime import datetime
 
+import uvicorn
 from fastapi import FastAPI
 from fastapi.responses import HTMLResponse
 
 from datastar_py.fastapi import DatastarStreamingResponse, ServerSentEventGenerator
+from datastar_py.starlette import sse_generator
 
 app = FastAPI()
 
@@ -15,7 +17,7 @@ HTML = """\
 		<head>
 			<title>DATASTAR on FastAPI</title>
 			<meta name="viewport" content="width=device-width, initial-scale=1.0"/>
-            <script type="module" src="https://cdn.jsdelivr.net/gh/starfederation/datastar/bundles/datastar.js"></script>
+            <script type="module" src="https://cdn.jsdelivr.net/gh/starfederation/datastar@v1.0.0-beta.11/bundles/datastar.js"></script>
 			<style>
             html, body { height: 100%; width: 100%; }
             body { background-image: linear-gradient(to right bottom, oklch(0.424958 0.052808 253.972015), oklch(0.189627 0.038744 264.832977)); }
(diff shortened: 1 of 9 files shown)


bencroker:
Any update on this, @gazpachoking ?

gazpachoking:
Still thinking on this. Updated it with some of the changes that have happened in the mean time. I guess there are 2 main bits to this that I'd welcome other people's opinions on:
  - A decorator that allows making your route a generator function directly, rather than having to define the generator separately than the route handler.
  - Allowing the sse iterables to just return fragments or dictionaries which automatically get interpereted as merge-fragments and merge-signals with default options without explicitly having to wrap them in `ServerSentEventGenerator.merge_fragments/signals`

These two pieces could be implemented separately. The second one could be made default for all frameworks without having to use any sort of decorator. Is that too much magic? The first one would use a decorator specific to each framework which would just wrap a generator function in the appropriate response class for that framework.

This shows off both ideas at once:
https://github.com/starfederation/datastar/blob/91551039a18183a03f765cdc390090b54fd42777/examples/python/fastapi/app.py#L86-L96

jmoppel:
My instinct on this is to wait, since the decorator is a convenience thing and not a must-have. Since we're unsure, let's build more things with Datastar and see where it goes. If we still think it is a good idea a bit down the road, we can always add this sort of thing later. There is no cost to the SDK to wait, since users still have a way to do the same thing as the decorator accomplishes.

gazpachoking:
What do you think about the implicit merge fragments and signals returns? Those could be implemented completely separate from the decorator

gazpachoking:
Okay, I split the implicit merge fragments/signals away from the decorator. The implicit wrapping now works with any of the frameworks without any decorator, and we can consider the two different ideas in this PR separately. Main post updated to reflect this as well.

gazpachoking:
I think idea 1 fits quite well with the behavior in v1 of doing a default merge-fragments and merge-signals for html and json content types. This would make similar assumptions, but on the server side, and you don't lose the benefits of SSE.

gazpachoking:
Splitting this out into separate PRs, the first one is up #922. Going to close this one now.

PR #811, opened 2025-03-31
Python SDK: Remove mixin from response classes that added SSE-generator functions

gazpachoking:
In #770 I changed the response classes to mix in the methods from `ServerSentEventGenerator`, with the idea that it saves an import and with one class lets you create a response and generate the appropriate events. After some usage and more thinking, I changed my mind on that being a good idea. Leaving everything using `ServerSentEventGenerator` makes everything more consistent between frameworks (both between the different python ones, and the ones in other languages.)

This would be breaking if anyone switched to the new way, so just leaving it as is would also be fine.


ajaegle:
+1 for the split into SSEG and DatastarStreamingResponse. It feels a bit weird to not have the `merge_fragments` on the SSEG like it is with other python integrations and sdks. It would be great to have some stability in this API soon and I like the version proposed here.

The currently released version (v0.4.3) is still using that single `responses.py` module that relies on a wrong/outdated import from `fasthtml.responses` instead of `fasthtml.starlette` which renders it unusable.

gazpachoking:
@lllama This one have your blessing?

gazpachoking:
Fixed up the merge conflicts, I think this is GTG again.

bencroker (approved):

bencroker:
Merged, thanks!

PR #814, opened 2025-04-01
Fix responsive layout for mobile devices

nishan-singh:
Added media queries for smaller devices using tailwind to avoid overflowing on content.

- The Input examples now split down into 2 lines on smaller devices

> `Output: 
    InputBox
`
- And the buttons to save or rest now go to the bottom right of the next line.
- The list (essays, memes and bundler) now has a  `right: 0` property, so it does not overflow.

### DIFF
--- a/site/shared.templ
+++ b/site/shared.templ
@@ -279,7 +279,7 @@ templ sidebarPageContents(sidebarGroups []*SidebarGroup, current *SidebarLink, c
 		<aside class="flex-col hidden gap-4 px-4 py-8 overflow-y-auto lg:flex min-w-80 bg-base-300 lg:visible">
 			@SidebarContents(sidebarGroups, current)
 		</aside>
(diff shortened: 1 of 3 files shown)


bencroker:
Thank you! A screenshot or two would make approving this much easier, if you wouldn’t mind.

nishan-singh:
BEFORE:
image
image
image
image
image

AFTER:
image
image
image
image
image

These are the changes that I made.


There was some padding on the main content
image

I decreased it because it was causing some overflow on my laptop

The content is already centered for wider screens
image

bencroker:
I haven’t forgotten about this, however, we‘re currently giving the site a redesign so I’m two minds about merging this. Thanks for your effort, I’ll ensure that the redesign is responsive and refer back to your improvements.

nishan-singh:
Thanks for letting me know!

PR #816, opened 2025-04-04
Clarify setting non-string values with data-signals

gazpachoking:
I was doing some silly stuff without realizing it isn't valid html like `<div data-signal-foo=false></div>` Added an explicit callout in the docs to use the object form when you need to set a signal to a non-string value.


gazpachoking:
Aborting this, I just realized that the non-object form is setting the signal with an *expression*

PR #817, opened 2025-04-07
Clojure SDK: Clj-kondo config - ring version bump - refactored a test

JeremS:
This PR contains 3 little fixes:
- a proper Clj-kondo config has been added to the SDK. This fixes warnings that the linter showed when it shouldn't have.
- the ring version has been bumped which led me to discover that the SDK didn't  handle the keep-alive header properly. This header is now properly added for connections using HTTP versions older than `1.1`.
- One test had some concurrent code that was a bit more complicated than was needed. This test now uses a simple promise instead of a count down latch.

No breaking change here since no public API has been touched, the changelog for the SDK has been updated.

Cheers,


bencroker (approved):

bencroker:
Merged, thanks!

PR #822, opened 2025-04-10
Python SDK: change `merge_fragments` to take a string rather than list

gazpachoking:
The other sdks, and the sse event docs have the merge fragments event type taking a string with the fragments. The python SDK was taking a list of strings. Everything worked fine, since it just concatenated them, but it meant fragments always had to be wrapped in a list, and made the python sdk not have parity with the others.

You can see with the changes to the examples in this PR that none of them were actually using a list, it was just a needless wrapping around the string. This is a breaking change, but one I think we should just bite the bullet and get in now.

### DIFF
--- a/examples/python/django/ds/views.py
+++ b/examples/python/django/ds/views.py
@@ -52,7 +52,7 @@ async def updates_asgi(request):
     async def time_updates():
         while True:
             yield DatastarStreamingHttpResponse.merge_fragments(
-                [f"""<span id="currentTime">{datetime.now().isoformat()}"""]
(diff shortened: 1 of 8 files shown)


bencroker:
Agreed. Thoughts, @lllama?

lllama (approved):

lllama:
LGTM :shipit:

gazpachoking:
Oops, I have to fix the override for fasthtml. Will do that in a little bit

gazpachoking:
Okay, fixed fasthtml as well now.

bencroker (approved):

bencroker:
Merged, thanks!

PR #823, opened 2025-04-10
Export to JSR

gavriguy:
This is still a draft.

I was able to export the project to JSR here https://jsr.io/@gavriguy/datastar-sdk@0.0.3
The user then need to import the node or the web implementation as follows:
`import { WebServerSentEventGenerator as ServerSentEventGenerator } from "@gavriguy/datastar-sdk";`

For now it uses my own username as I don't want to register datastar namespace without an approval 

To publish the package  to JSR all is needed is to run `pnpm dlx jsr publish  --allow-slow-types`

I wanted to keep the ` --allow-slow-types` as i didn't want to change the codebase in this PR. But if the PR is approved we can change the codebase so it won't need the   ` --allow-slow-types` flag

### DIFF
--- a/sdk/php/README.md
+++ b/sdk/php/README.md
@@ -30,10 +30,12 @@ use starfederation\datastar\enums\EventType;
 use starfederation\datastar\enums\FragmentMergeMode;
 use starfederation\datastar\ServerSentEventGenerator;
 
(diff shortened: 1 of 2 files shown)


bencroker:
I converted the PR to a draft. Let me know once you]re further along. Please also base it against the `develop` branch.

bencroker:
Any update on this, @gavriguy ?

gavriguy:
I will open a new one on develop

PR #826, opened 2025-04-11
Externalize interface compatibility for two optional template engines

dkotik:
Templ and GoStar are currently module dependencies for Golang SDK. They should not be required for 1.0 release. This step replaces two component interfaces with copies that ensure compatibility without having to include various optional engines as dependencies with each Datastar deployment.

Also, SDK should become an independent module and have its own go.mod file. That is a more drastic future change that will be made easier by this pull request.

### DIFF
--- a/sdk/go/fragments-sugar.go
+++ b/sdk/go/fragments-sugar.go
@@ -1,10 +1,10 @@
 package datastar
 
 import (
+	"context"
 	"fmt"
+	"io"
 
-	"github.com/a-h/templ"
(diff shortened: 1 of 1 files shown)


dkotik:
Alternatively, the site folder might get its own go.mod file instead. This will preclude having to move the SDK mod location.

delaneyj:
This is a great idea!

delaneyj (approved):

bencroker:
Merged, thanks!

PR #827, opened 2025-04-11
Update comments and documentation for Golang SDK

dkotik:
Added comments to `execute.go` file. I will make similar changes to
other files. I submit those as a way of getting feedback.

The only public API change is the removal of execute script options
struct. It is changed to private by letter case because the options
pattern is used to configure script execution. The struct serves no
purpose and pollutes the public API.


delaneyj (approved):
Thank you so much!

bencroker:
Merged, thanks!

PR #828, opened 2025-04-11
Use go tools 

alexbozhenko:
Yo!
Have you guys heard of [reproducible builds](https://reproducible-builds.org/docs/commandments/)?
How about [hermeticity](https://bazel.build/basics/hermeticity#overview) ?

> go install github.com/a-h/templ/cmd/templ@latest

What? [3 weeks later version will change from 14.5.75 to 15.1.200](https://x.com/htmx_org/status/1899917413153481070), and it will stop building.

Luckily, starting with go 1.24, go can manage tools dependencies, and record them in go.mod:
https://go.dev/doc/modules/managing-dependencies#tools

So I propose to use that, and rely on awesome go tooling to manage the installation.
```
# go tool task templ
task: [templ] go tool templ generate .
(✓) Complete [ updates=183 duration=190.490912ms ]
```

This is just Proof of Concept. Let me know if you are open, and I will update other `go install`s in the taskfile

UPD: follow-up in https://github.com/starfederation/datastar/pull/829

PR #829, opened 2025-04-12
Fix go tools

alexbozhenko:
The presence of a link to htmx meme in https://github.com/starfederation/datastar/pull/828 tricked experienced authors of this repo, so they thought I knew what I was doing, and they rushed to merge the PR.

So this is a follow-up. In this PR, we:
* update go to 1.24, to actually allow us to use `go tool ...`
* make qtc a go tool
* remove tools installation from dockerfile dev

Not sure about managing the task itself as go tool, since, depending on the CI setup, it may take extra time to build it. But I guess we will see it in practice...

## Reproducible test plan
I run the following to verify that nothing is broken:
```
echo "build  image-build  Makefile test" | xargs -n1  make
```
And I also run commands outside of container:
```
task tools
task build
task 
task test
```

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -53,7 +53,7 @@ Below are details and options available in each of the provided `make` commands:
 
 You can pass in optional CLI arguments to override the default settings Datastar dev uses:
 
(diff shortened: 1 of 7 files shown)


alexbozhenko:
Tests are still failing. 
To make them pass: 
1. land this PR
2. run https://github.com/starfederation/datastar/actions/workflows/build-docker-images.yml
3. run tests again

delaneyj (approved):

bencroker:
Merged, thanks!

PR #831, opened 2025-04-12
Add start history to the readme

alexbozhenko:
can hide under if do not want to waste space
<details></details>


bencroker:
Thanks for the suggestion. The readme is auto-generated, so your code change will be overwritten with the next build. I’ll consider adding it separately.

PR #834, opened 2025-04-12
Add hot reload example to Golang SDK

dkotik:
Updated the Golang SDK README.md file to include a list of examples that
will be expanded in the future. A basic example is moved into its own
directory. A hot reload example is added alongside it.


delaneyj (approved):

bencroker:
Merged, thanks!

PR #835, opened 2025-04-12
chore: update go modules in go example

alexbozhenko:
Also, fix tests. Follow up to 
https://github.com/starfederation/datastar/pull/828
https://github.com/starfederation/datastar/pull/829


alexbozhenko:
Ok, tests finally passed, and test run time was about the same, so using `go tool task` should work.

bencroker:
Thanks!

PR #837, opened 2025-04-14
Python SDK: Allow objects with the __html__ protocol to be used for merge_fragments

gazpachoking:
There are multiple libraries (fasthtml, htpy, django, jinja) that use the standard of having an `__html__` method on objects that can render to html. This adds support for passing those objects straight to merge_fragments, without having to stringify them first. Added benefit is that the special casing of fasthtml can go away and it just becomes another starlette framework.

### DIFF
--- a/sdk/python/src/datastar_py/fasthtml.py
+++ b/sdk/python/src/datastar_py/fasthtml.py
@@ -1,16 +1,2 @@
-from typing import override
-
-from fastcore.xml import to_xml
-
-from .sse import SSE_HEADERS, ServerSentEventGenerator
-from .starlette import DatastarStreamingResponse as _DatastarStreamingResponse
-
-
-class DatastarStreamingResponse(_DatastarStreamingResponse):
-    @classmethod
-    @override
-    def merge_fragments(cls, fragments, *args, **kwargs):
-        if not isinstance(fragments, str):
-            fragments = to_xml(fragments)
-        # From here, business as usual
-        return super().merge_fragments(fragments, *args, **kwargs)
+from .sse import ServerSentEventGenerator
+from .starlette import DatastarStreamingResponse
--- a/sdk/python/src/datastar_py/sse.py
+++ b/sdk/python/src/datastar_py/sse.py
@@ -1,6 +1,6 @@
 import json
 from itertools import chain
-from typing import Optional
+from typing import Optional, Protocol, Union, runtime_checkable
 
 import datastar_py.consts as consts
 
@@ -9,6 +9,12 @@ SSE_HEADERS = {
     "Connection": "keep-alive",
     "Content-Type": "text/event-stream",
 }
+
+
+@runtime_checkable
+class _HasHtml(Protocol):
+    def __html__(self) -> str:
+        ...
 
 
 class ServerSentEventGenerator:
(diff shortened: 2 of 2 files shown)


jmoppel:
I like the idea. A few suggestions:

* Type it as a class and put in the doc string and docs that it must have an `__html__` dunder method or attribute. Throw an error if it doesn't have an `__html__` method.

* It would be better to type these things per library than going with some new type that no one will understand. In Django it would be a `SafeString` for example and other frameworks will have their own names..

lllama:
This seems to make sense. Doesn't appear to affect any of the existing functionality (and it'll play nice with Sanic as that uses Jinja, so that's a win for me 🤣)

I think it already addresses @jmoppel's comments unless I've misunderstood `Protocol`. I don't think we need a per-framework thing as the expectation is that we're being passed validation html as we're more at the transport layer.

axelknock:
This is a good solution. For the FastHTML response, [a call to `__html__` is just a call to `to_xml`](https://github.com/AnswerDotAI/fastcore/blob/main/fastcore/xml.py#L212) anyway.

gazpachoking:
> * Type it as a class and put in the doc string and docs that it must have an `__html__` dunder method or attribute. Throw an error if it doesn't have an `__html__` method.

Not quite sure what you mean. The way I did it IDEs and type checkers will understand that both strings and anything with a `__html__` method is allowed to be passed in to the merge_fragments helper.

> * It would be better to type these things per library than going with some new type that no one will understand. In Django it would be a `SafeString` for example and other frameworks will have their own names..

I agree it's unfortunate we don't have a common name for ClassThatCanProduceHtml, which is why I went with HasHtml. Type checkers and IDEs understand the way I've done it natively, but humans may have to look closer to understand what HasHtml is. (I did add a docstring for humans now though.) I think it's much nicer to have it in this generic form, because you can use any tool that generates html (and follows this convention) with any framework. For example htpy is not framework specific, you can use it with any of the python web frameworks. Safestring from django doesn't actually need this, because it's already a string. In that case implementing `__html__` is designed to tell other html _producing_ tools that they shouldn't escape the text it produces when embedding it.

jmoppel:
> I think it already addresses @jmoppel's comments unless I've misunderstood `Protocol`. I don't think we need a per-framework thing as the expectation is that we're being passed validation html as we're more at the transport layer.
...
> Not quite sure what you mean. The way I did it IDEs and type checkers will understand that both strings and anything with a __html__ method is allowed to be passed in to the merge_fragments helper.

I haven't needed to use them, but I think you're both correct, and I was mistaken. I think Protocols would work in the generic case.

What does everyone think about making the name more explicit? HasHtml could mean several things. Subjectively, something like `HasDunderHtml`, `HasHtmlMagic`, `HasHtmlMagicMethod`, etc. are all more obvious to my eye (the last one is rather long tho).
   
I hold that it would be much better to make available the idiomatic language and types for each framework. I confess that I didn't know that SafeString even had an `__html__` method until I looked it up. That method is only used for compatibility with other libraries/templating engines, and isn't something that comes up when you're simply building an app. I doubt most Django devs would know to use SafeString in this instance. Note: this could easily be a follow on commit.

gazpachoking:
> What does everyone think about making the name more explicit? HasHtml could mean several things. Subjectively, something like `HasDunderHtml`, `HasHtmlMagic`, `HasHtmlMagicMethod`, etc. are all more obvious to my eye (the last one is rather long tho).

Naming is hard. I'm open to different names, though I'm not sure I like these better. `HtmlProducer`?

> I hold that it would be much better to make available the idiomatic language and types for each framework. I confess that I didn't know that SafeString even had an `__html__` method until I looked it up. That method is only used for compatibility with other libraries/templating engines, and isn't something that comes up when you're simply building an app. I doubt most Django devs would know to use SafeString in this instance. Note: this could easily be a follow on commit.

I disagree. The idea here is to be able to use any html producing tool with any framework, even tools that don't exist right now so would be impossible to document. Not something that's specific to any specific framework. I don't think django has anything that applies here, as stated before SafeString is a string, so works fine before this PR.

gazpachoking:
> I don't think django has anything that applies here, as stated before SafeString is a string, so works fine before this PR.

I take that back, I just had a peek at django. `Form`, `FormSet`, `BoundField`, `ErrorList`, `ErrorDict` all  follow this standard. You could pass a `Form` instance into merge_fragments with this PR and it would just work.

bencroker:
So, is this PR ready for prime-time?

gazpachoking:
I think we are good to go. I was just tweaking naming and comments, which don't affect the user api. I'm happy with them now. :shipit:

bencroker (approved):

bencroker:
Merged, thanks!

PR #838, opened 2025-04-14
Expand Golang SDK documentation by adding comments to every method.

dkotik:
Added a comment to every public method. Removed inaccessible options
structs from public API by letter case change. Exposed one hidden option
to public API. Refactored fragment merge mode parsing and added a
matching validation test.


delaneyj (approved):
Awesome, thank you!

bencroker:
Merged, thanks!

PR #839, opened 2025-04-14
Remove panic call from one ExecuteScriptOption (public API change)

dkotik:
Changed the shape of the ExecuteScriptOption option to return an error,
which allows validation of the Option parameters. This allowed removing
a panic call from one of the options.

Panic calls are common in the standard library. However, Datastar calls
return errors. Mixing panics and error returns in the same construct is
slight code smell. More importantly, the mixing produces a less
intuitive and predictable API.


dkotik:
I would like to similarly change the shape of other option calls if this change goes through.

delaneyj:
No, there is :100: a server configuration issue and a panic, not a runtime error.  WON'T FIX

PR #842, opened 2025-04-15
Add missing WithMergeFragmentsEventID option

dkotik:
The merge fragments call configuration struct had an EventID field and logic that added it to the SSE call, but no option to configure it. Added that missing option.


delaneyj (approved):
Awesome, thanks!

PR #844, opened 2025-04-15
Add `X-Accel-Buffering` to Python SDK

Laurent2916:
to prevent nginx from buffering stuff

### DIFF
--- a/sdk/python/src/datastar_py/sse.py
+++ b/sdk/python/src/datastar_py/sse.py
@@ -8,6 +8,7 @@ SSE_HEADERS = {
     "Cache-Control": "no-cache",
     "Connection": "keep-alive",
     "Content-Type": "text/event-stream",
+    "X-Accel-Buffering": "no",
 }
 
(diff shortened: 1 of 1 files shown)


bencroker:
Seems like a good idea. @lllama ?

bencroker:
Still waiting on a review from @lllama...

lllama:
Hmm. I think there was discussion on Discord that we'd avoid adding specific headers in the SDKs. Do we know if the other SDKs are adding this?

bencroker:
The PHP SDK uses this as a default header, and the Ruby one too, so I think it depends on how common it is to use nginx with Python.

lllama:
Still on the fence about this one but I don't think it'll do any harm or introduce surprising behaviour, so ok to merge.

bencroker (approved):

bencroker (approved):

bencroker:
Merged, thanks!

PR #849, opened 2025-04-17
update version to 0.1.0-beta.1 for rust sdk

GlenDC:
ready for publish

- changelogs per sdk's is not done it seems, so I did not add that
- added no rust specific QA GH action as it doesn't seem to be done for other SDK's and so I imagine we want to keep things simple and open.

I ran `clippy` and other rust tools locally and all seems ok.

Once this is merged we should be able to run

> cargo publish

and all should be ok

### DIFF
--- a/sdk/rust/Cargo.toml
+++ b/sdk/rust/Cargo.toml
@@ -12,7 +12,7 @@ license = "MIT OR Apache-2.0"
 name = "datastar"
 readme = "README.md"
 repository = "https://github.com/starfederation/datastar-rs"
-version = "0.1.0"
+version = "0.1.0-beta.1"
 rust-version = "1.85.0"
 
 [dev-dependencies]


GlenDC (commented):
Ready for merge if you have it @bencroker .

bencroker:
Will `cargo.io` accept `1.0.0-beta.1` as a valid version? That’s what most SDKs are using for versioning.

Were there no other changes you wanted to make before publishing?

GlenDC:
> Will cargo.io accept 1.0.0-beta.1 as a valid version? That’s what most SDKs are using for versioning.

Yes it would.

I assume you ask because you wanna keep the versioning the same as the Js or at least indicate a relationship with that 1.0 thing. However if you really wanna be semver compliant than I don't think you'll be able to ever keep the versions between all the different sdk's and js lib in sync. If you do want the semver-compliance AND you want to keep the versions in sync it would mean that a breaking change in _any_ sdk would require _all_ projects to bump major as well.

Now here comes the thing. Projects like `Axum` and `Rama` (two of the three http servers exposed by this rust SDK) are not yet on `1.x.y` themselves yet. This doesn't mean these projects introduce breaking changes all the time or are unstable or not production ready. It just means that they still still are open to some breaking changes in future and do not want to commit yet to a longterm major release. Because if you have `0.x` as your version than your 'minor' is your 'major', at least in Rust world.

And here comes the crux of the story. If we would already go for `1.0.0-beta.x`, assuming you want to so because you want to soon go to `1.0.0` (in sync with the js lib) than we would need to bump to `2.0.0`, `3.0.0` each time we have a new major version of either Axum or Rama. Which is fine for me btw, but abnormal for most of the rust ecosystem.

> Were there no other changes you wanted to make before publishing?

No as the other stuff was not code related but more about CI, a high level hello test and changelogs. But the test was already there, per-sdk-specific changelogs we do not do it seems and there are also no other QA-like (e.g. linters / formatters / etc... ) actions for other sdk's, so that also got dropped from my plans.

Once I start using the sdk in our company projects I might require some updates on this crate, which is why the `-beta.x` part is there. But so far it looks good. Already tried it for a quick test and it works fine for what I tried it for. Only bug I found was apparently a bug in the JS side of things but according to the discord server that was already fixed and to be shipped in the next release of the bundle.

bencroker (approved):

bencroker:
Thanks for the explanation, merged!

PR #850, opened 2025-04-17
Python SDK: Add read_signals helpers

gazpachoking:
Adds read_signals functions to all the supported python frameworks. Also adds a dependency for FastAPI so that they can be read idiomatically in that framework. @jmoppel Can you validate the Django one?

I also put a ruff config and formatted the files. Added `__all__` to the files as well to make sure that imports that were unused and only included for re-exporting don't get accidentally removed. These can be separated out into different PRs if needed, I was just wanted the ability to easily lint and format my work and ended up adding it in here.


gazpachoking:
Thoughts on the fastapi dependency name? It's `SignalsDep` right now

gazpachoking:
Don't think the test failures are anything to do with this change..

gazpachoking:
I changed the FastAPI dep to be called `ReadSignals` rather than `SignalsDep`. I think it's a bit better.

gazpachoking:
I tested the django read_signals helper now, and changed the behavior to return None rather than error if the request was not sent from datastar, or had no signals attached. This lets FastAPI and Litestar use read_signals as a dependency, even if the route handler gets hit by requests that aren't from datastar. I think this is in a pretty good state now.

Also added examples of how to use read_signals for all the different frameworks.

bencroker:
> I changed the FastAPI dep to be called `ReadSignals` rather than `SignalsDep`. I think it's a bit better.

Naming conventions should follow the SDK spec as closely as possible, so it should absolutely be `ReadSignals` or `read_signals`, as appropriate.
https://github.com/starfederation/datastar/tree/develop/sdk#readsignalsr-httprequest-signals-any-error

gazpachoking:
Previous implementation raised an exception with the `excludeSignals` option coming up in the next RC. Fixed that issue, and factored out the logic to one place so that it's not implemented separately for each framework.

bencroker:
This is looking great! If we could get a Python SKD contributor to look at and approve this PR, that would be reassuring, as I’m less familiar with Python.

gazpachoking:
In hopes of making reviewing this easier I stripped out all the other fixes and cleanups not related directly to read_signals. I still think those are good, but I'll raise them in a separate PR.

gazpachoking:
@lllama Mind taking a peek here?

gazpachoking:
Oh, I guess one pertinent question I didn't consider is where this should go. I have it in the sse file because that's where our common stuff has gone, but really it doesn't have anything to do with sses.

EDIT: I guess the bit in sse.py is private, so it doesn't really matter. Maybe better to move that up to `__init__.py` though. All the user facing stuff is in the framework modules.
EDIT2: Went ahead and did that. It's private, so it can always be moved later with no impact. But having it in sse.py didn't seem right.

lllama (approved):

bencroker (approved):

bencroker:
Merged, thanks!

PR #851, opened 2025-04-17
feat(plugins/indicator): keep handling events until cleanup

jonnelaaglandwinder:
Fetch indicators currently only work once, so triggering a new fetch from the same element won't indicate its progress.

I changed the indicator behaviour to keep watching until cleanup, rather than only triggering once. This is achieved by tracking orphaned watchers using a WeakSet; orphaned watchers check the set after handling the event, disconnecting from the DOM if they find themself in the set. This may cause regressions for anything relying on the old behaviour, but the new behaviour is probably more in line with what one would expect.
This still isn't perfect since triggering a new fetch while another one is in flight may cause the signal to return to false too early, but at least indicators now keep updating until removed.


bencroker:
Thanks for the PR, however, this has already been fixed (in a more robust way) in the current release candidate (development of which is happening in a private repo).

mstg:
@jonnelaaglandwinder Thank you, currently using this as a quick workaround!

PR #854, opened 2025-04-18
improve quality of rust sdk and add tooling

GlenDC:
also bump version to 0.1.2

- add makefile with all commands for rust sdk developers to run all relevant checks and test, which also auto formatting of code and toml deps
- fixed broken clippy rules (this also required fixing some stuff in consts)

Also fixed the docs.rs attributes to ensure _all_
modules and symbols get published in docs,
instead of only those that are available in default features (which was missing the framework specific stuff in docs)

Only nitpick issue left: in generated code (consts.rs) some descriptions can contain 'html' code which is not escaped, pretty certain this html will be rendered in docs.rs as rust docs renders docs as markdown (which includes html), but given that's auto generated I have left that for now as-is.


bencroker (approved):

bencroker:
Merged, thanks!

PR #856, opened 2025-04-20
[sdk-rust ; 0.2.0] remove need for '.into()' => DatastarEvent

GlenDC:
this is a breaking change however as existing code using '.into()' will no longer compile as the compiler wouldn't know what is being into'd. As such I bumped major version to 0.2.0

the resulting library is however a 'bit' easier to use

For cases where you do need to return multiple different kinds of datastar events I added convenient methods to easily turn such fragments into datastar events (.into_event())

### DIFF
--- a/sdk/rust/Makefile
+++ b/sdk/rust/Makefile
@@ -2,18 +2,19 @@
 
 all:
 	@echo "Usage:"
-	@echo "fmt           - run the rust formatter"
-	@echo "sort          - sort TOML dependencies"
-	@echo "lint          - combine fmt+sort"
-	@echo "check         - shallow check of rust code (pre-compile)"
-	@echo "clippy        - run clippy checks"
-	@echo "doc           - doc checks"
-	@echo "hack          - test feature matrix compatibility"
-	@echo "test          - run all unit and doc tests"
-	@echo "qa            - combine lint+check+clippy+doc+hack+test"
-	@echo "hello-axum    - run hello-world example using the Axum framework"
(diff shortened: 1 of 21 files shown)


GlenDC (commented):
Self-reviewed 2 times, all looks good to myself.
But open to feedback of course.

sdk/rust/Makefile
```diff
@@ -2,18 +2,19 @@
...
+	@echo "fmt                  - run the rust formatter"
+	@echo "sort                 - sort TOML dependencies"
+	@echo "lint                 - combine fmt+sort"
+	@echo "check                - shallow check of rust code (pre-compile)"
+	@echo "clippy               - run clippy checks"
+	@echo "doc                  - doc checks"
+	@echo "hack                 - test feature matrix compatibility"
+	@echo "test                 - run all unit and doc tests"
+	@echo "qa                   - combine lint+check+clippy+doc+hack+test"
+	@echo "detect-unused-deps"  - detect unused deps for removal"
```
gazpachoking: Extra quote here?

GlenDC: Nice spot, fixed it.

GlenDC:
Applied feedback, updated deps while I was at it and have also ensured that examples are qa'd as well.

FYI @bencroker self-reviewed it once again, lgtm.
Might be useful if @jmstevers also gives their blessing.

ghost:
Yep, LGTM

bencroker (approved):

bencroker:
Merged, thanks all!

PR #859, opened 2025-04-21
Fix to README, including version so it doesn't default to 0.2.x of the library

peteyako:
Seems to default to 0.2.x if version isn't specified.


delaneyj (approved):

bencroker:
Merged, thanks!

PR #861, opened 2025-04-23
Fix label for middle download time stat

ElijahSink:
"Fast 4G" < "Slow 3G" < "3G" doesn't make sense. It only makes sense if the middle stat label was intended to be "Slow 4G"


bencroker (approved):

bencroker:
Merged, thanks!

PR #864, opened 2025-04-24
Add working log action plugin to the README

justinmoon:
Would be better to use `type: PluginType.Action` but I don't know how to import that from the CDN

Some context in [this Discord thread](https://discord.com/channels/1296224603642925098/1364842481006673930)


bencroker:
Thanks for the PR! If you’re going to show a sample plugin, please consider adding comments to the code explaining what it does.

PR #871, opened 2025-04-29
Move datastar package files from `sdk/go` to `sdk/go/datastar`

cshuman:
…ith the package name to avoid needing to use the datastar alias when importing.


delaneyj (approved):

cshuman:
@delaneyj looks like there's a merge conflict because the README.md now has the pre-release tag. Do you want to keep that in the README or remove it to prepare for v1?

sdk/go/README.md
bencroker, line 1: This file is auto-generated by the build process. Please also modify the existing `consts_go.qtpl`.

bencroker:
> @delaneyj looks like there's a merge conflict because the README.md now has the pre-release tag. Do you want to keep that in the README or remove it to prepare for v1?

Let’s keep it for now and change when v1 launches.

cshuman:
merge conflict fixed in README and fixed the location for the consts.go file.

sdk/go/README.md
cshuman: to clarify, it's not the README that is auto-generated, right? I think I found what needs to be updated in the build/run.go file so that the consts.go file gets put in the right place. thank you!

bencroker: The readme is auto-generated from https://github.com/starfederation/datastar/blob/develop/build/consts_go.qtpl

cshuman: it looks like [consts_go.qtpl](https://github.com/starfederation/datastar/blob/develop/build/consts_go.qtpl) generates the [consts.go](https://github.com/starfederation/datastar/blob/develop/sdk/go/consts.go) file.
When you say that the readme is auto-generated, do you mean the readme at the root of the repo? it looks like that readme is generated from [consts_datastar_readme.qtpl](https://github.com/starfederation/datastar/blob/develop/build/consts_datastar_readme.qtpl)


Sorry I wasn't clearer in my first message about the readme. I was referring to the readme in the root of the sdk/go directory [here](https://github.com/starfederation/datastar/blob/develop/build/consts_datastar_readme.qtpl). 

*That* readme has the following:

`go get -u github.com/starfederation/datastar/sdk/go@v1.0.0-beta.11`

which is helpful to bring in the latest go sdk/module of the v1 pre-release/beta tag instead of the latest v0 tag. Once you have a proper v1 you'll want to remove that version tag to get the latest (non-pre-release) v1 version.

bencroker: Sorry, I was being an idiot!

bencroker:
Merged, thanks!

PR #872, opened 2025-04-29
Sdk/rust improvements

GlenDC:


build/consts_rust.qtpl
bencroker, line 1: The request header should be `Datastar-Request`, see https://github.com/starfederation/datastar/blob/develop/build/consts_datastar_client.qtpl#L10

GlenDC: Headers are case insensitive. And for h2 and beyond uppercase is forbidden. Some rust libs can do difficult about static header strings with uppercase

bencroker (approved):

bencroker:
Merged, thanks!

PR #869, opened 2025-04-28
create a separate go module for the site builder

dkotik:
Completes pull request #826 by removing unnecessary dependencies for
optional template engines.

The module was named `sitebuilder` instead of `site` because the latter
creates an ambigous import. It can be fixed later, after the SDK import
is bumped to the latest version.

PR #874, opened 2025-04-30
Add new `go.mod` file

dkotik:
Go module path is moving to a new path. At this time, it is best to create a new go.mod file. This also removes the need to provide alternative name for import.

Related:

- https://github.com/starfederation/datastar/pull/869
- https://github.com/starfederation/datastar/pull/871


dkotik:
this will only get fixed after a tagged version is released, so module can be installed from the new path:

```sh
site/smoketests/merge_options_test.go:7:2: no required module provides package github.com/starfederation/datastar/sdk/go/datastar; to add it:
	go get github.com/starfederation/datastar/sdk/go/datastar
```

bencroker:
We have this in the new site, so I’ll close here. Thanks for the PR!

PR #876, opened 2025-05-02
Python SDK: Stop formatting merge-signals events for humans

gazpachoking:
Merge signals was adding newlines and indents, causing unneeded overhead when updating signals. Just send it all on one line now.
image

### DIFF
--- a/sdk/python/src/datastar_py/sse.py
+++ b/sdk/python/src/datastar_py/sse.py
@@ -114,9 +114,8 @@ class ServerSentEventGenerator:
         if only_if_missing:
             data_lines.append(f"data: {consts.ONLY_IF_MISSING_DATALINE_LITERAL} true")
 
-        data_lines.extend(
-            f"data: {consts.SIGNALS_DATALINE_LITERAL} {signalLine}"
-            for signalLine in json.dumps(signals, indent=2).splitlines()
+        data_lines.append(
+            f"data: {consts.SIGNALS_DATALINE_LITERAL} {json.dumps(signals)}"
         )
 
         return ServerSentEventGenerator._send(


lllama:
Not sure about this one. Any overhead is going to be minimal and it makes inspecting events a lot easier.

bencroker (approved):

bencroker:
I tend to agree with @gazpachoking. SDKs shouldn’t be reformatting event data for human readability.

Merged, thanks!

lllama:
Well technically it's not reformatting at all: we need a string and so it's producing a string. Just adding some whitespace while it does it.

But no big deal either way.

bencroker:
My understanding was that it was reformatting the value passed into `merge_signals`, for example `ServerSentEventGenerator.merge_signals({"currentTime": f"{now}"})`.

bencroker:
Point taken – adding whitespace (formatting) is what I meant by reformatting.

gazpachoking:
Just some examples of the size difference of the event with the 2 methods for posterity. I wouldn't call that amount of overhead minimal (x1.5 and x2.3 in these examples), but I agree it doesn't matter too much because the bandwidth from merge-signals events is probably going to be small no matter what in the grand scheme of things.
```
event: datastar-merge-signals
retry: 1000
data: signals {
data: signals   "time": 130
data: signals }

event: datastar-merge-signals
retry: 1000
data: signals {"time": 130}

len(old_a)=103
len(new_a)=71

len(old_a)/len(new_a)=1.4507042253521127

event: datastar-merge-signals
retry: 1000
data: signals {
data: signals   "foo": "blah blah",
data: signals   "bar": "baz",
data: signals   "x": [
data: signals     1,
data: signals     2,
data: signals     3
data: signals   ]
data: signals }

event: datastar-merge-signals
retry: 1000
data: signals {"foo": "blah blah", "bar": "baz", "x": [1, 2, 3]}

len(old_b)=244
len(new_b)=108
len(old_b)/len(new_b)=2.259259259259259
```

lllama:
As spake by the CEO: can a man truly speak of overhead when he has not measured in the real world where one's messages are beset upon by the daemons of compression and JITing runtimes?

lllama:
But tbh I use Neovim (btw) so it's trivial to copy the message and wrangle as needed.

PR #878, opened 2025-05-03
Python SDK: Basic litestar support

gazpachoking:
Adds litestar support in the python sdk. I decided not to use Litestar's built in SSE response, since that requires a special sse event object to be returned from the iterator. Making our own class lets us just return the same ServerSentEventGenerator events that we do for all the other frameworks.

### DIFF
--- a/examples/python/litestar/app.py
+++ b/examples/python/litestar/app.py
@@ -1,0 +1,69 @@
+import asyncio
+from datetime import datetime
+from typing import AsyncGenerator
+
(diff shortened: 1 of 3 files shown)


2a5A1Ghu1:
Looks like litestar should be added here as well 

https://github.com/starfederation/datastar/blob/d24ae7402a7bc2482981a4a4d35cf45f8c97e4b3/sdk/python/pyproject.toml#L13

bencroker (approved):

bencroker:
Merged, thanks!

PR #879, opened 2025-05-04
add doctest for the rust sdk README example

GlenDC:
As we are dealing with multiple kind of events the usage of `into` is fine in this example,
alternatively is to use `into_event` but it would be the same in this case, `into_event` is more there
to make it also work for when your output type is `impl Into<DatastarEvent>`.


bencroker (approved):

bencroker:
Merged, thanks!

PR #883, opened 2025-05-09
Python SDK: Update README

2a5A1Ghu1:
Added supported framework url for Litestar and sorted frameworks in lexicographical order


gazpachoking:
Looks good. We should probably add Starlette as well

2a5A1Ghu1:
> Looks good. We should probably add Starlette as well

Done

bencroker (approved):

bencroker:
Merged, thanks!

PR #885, opened 2025-05-09
Fix bash test sort

ismasan:
There seems to be a bug in the `./sdk/test/normalize.sh` script, depending upon by `test-all.sh`

```
          >>>  close("sort", <<<
awk: illegal statement at source line 7 in function flush_data
awk: illegal statement at source line 7 in function flush_data
awk: syntax error at source line 7 in function flush_data
```

The cause is that there's a `close("sort", "to")` call, whereas AFAIK the `close()` function expects a single argument.

This PR removes that unnecessary second argument to `close()`, which breaks on my Mac.

I double-checked with @jmstevers and this second argument doesn't seem to be needed.


bencroker:
I’m not sure at what point you started this PR, but you were probably way behind `develop`. This PR includes changes to PHP files, so I’m not going to merge. Please reopen against the latest commit in the `develop` branch.

ismasan:
@bencroker 💩 not sure what happened there, apologies. Will re-work.

ismasan:
Here: https://github.com/starfederation/datastar/pull/886

Apologies again.

PR #901, opened 2025-05-19
Python SDK: Properly document the SSE event ID as a string

gazpachoking:
Python sdk was type hinting the sse event ID as an integer rather than a string.


bencroker (approved):

bencroker:
Merged, thanks!

PR #907, opened 2025-05-24
Remove view transition meta tag injection

Regaez:
The meta tag for enabling view transitions is deprecated; it was only used by old versions of Chrome when the experimental feature flag was required to enable view transitions. 

Per the [draft spec](https://drafts.csswg.org/css-view-transitions-2/), cross document view transitions should be enabled via the CSS rule `@view-transitions`.

Thus, I believe this datastar plugin should no longer append this element to the document head as it's mostly ineffective runtime code that isn't aligned with the spec. This should also have the added benefit of saving a few bytes from the bundle.


bencroker (approved):

bencroker:
Thank you!

PR #908, opened 2025-05-25
add tests to verify that #905 works as intended

### LINKED ISSUE #905: Rust SDK SSE events flushing bug
### Bug Report

I've noticed that the last SSE event is not actually send, and my page displays previous value of signal being updated by the backend (Rust axum app). When inspecting the response in the browser the last event was only partially send:

```
event: datastar-merge-signals
data: signals {current_round: 0}


event: datastar-merge-signals
data: signals {current_round: 13}


event: datastar-merge-signals
data: signals {current
```

After I changed the code to use a shorter `cur_round` signal name, the issue is gone. I looked around the `Sse` implementation, and since it's fairly low-level and non-trivial, my guess is that something about the way futures are pulled is incorrect and leads to some bytes getting stuck in the future, until next even is sent.

I'm using Rust crate `datastar v0.2.1` with Axum. Datastar version is `v1.0.0-beta.11`, but I don't think it has anything to do with the frontend part.

### Datastar Version

v1.0.0-beta.11

GlenDC:
All looks good. Only thing that I can think of what is happening for the user is that their chunk encoding is not working as intended, but that cannot be reproduced on my end.

Closes #905

### DIFF
--- a/sdk/rust/src/axum.rs
+++ b/sdk/rust/src/axum.rs
@@ -188,14 +188,16 @@ mod tests {
     use {
         super::Sse,
         crate::{
-            prelude::ReadSignals,
-            testing::{self, Signals, base_test_server},
+            prelude::{MergeSignals, ReadSignals},
+            testing::{self, Signals, base_test_server, test_server_merge_signal_complete},
         },
+        async_stream::stream,
         axum::{
             Router,
             response::{Html, IntoResponse},
             routing::{get, post},
         },
+        std::time::Duration,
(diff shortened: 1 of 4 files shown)


GlenDC:
Gonna close this PR as I don't think this is the right approach to this issue. I think #909 is the more appropriate way, until tan this might be observed. There's no control via this approach at this higher abstract level to force flushing or have influence over chunked encoding etc. The rust SDK is due to a heavy refactor, for better foundations, some optimizations and support.

PR #912, opened 2025-05-29
Django version bump and small docs update

avkoval:
Just small dependencies update and a command example for newcomers to run the example

### DIFF
--- a/examples/python/django/README.md
+++ b/examples/python/django/README.md
@@ -7,7 +7,7 @@ To take full advantage of Datastar, we recommend (but do not require) that you u
 
 ## ASGI (default)
 
-Install the dependencies, start runserver as usual, and go to http://127.0.0.1/ to see the demo.
+Install the dependencies, start runserver as usual, and go to http://127.0.0.1:8000/ to see the demo.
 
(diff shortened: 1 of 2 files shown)


examples/python/django/README.md
```diff
@@ -21,6 +21,8 @@ To use WSGI make sure runserver is stopped and make the following changes to the
 
 2. In `INSTALLED_APPS` comment out `"daphne"`,
 
-3. Start runserver and go to http://127.0.0.1/wsgi/ and watch the demo.
-
+3. Start runserver and go to http://127.0.0.1::8080/wsgi/ and watch the demo.
```
gazpachoking: Too many colons here.

gazpachoking (commented):
Looks like the default port is 8000 rather than 8080. Perhaps we should also show the uv run command up in the ASGI section and not just down under the WSGI section.

Unrelated to your changes, but while we are at it, maybe we should update the WSGI instructions to say the full name of the settings, i.e. WSGI/ASGI_APPLICATION rather than APP.

Other than that, seems good to me!

gazpachoking:
lgtm 👍

gazpachoking (approved):

bencroker (approved):

bencroker:
Merged, thanks!

PR #914, opened 2025-06-02
Set a default on_error handler to log exception backtrace

ismasan:
In the current versions, exceptions raised within streaming threads (ex. a templating error) are swallowed by the runtime, unless an explicit `on_error` handler is configured.

This PR adds a default error handler to log the exception and its backtrace to the available logger.

Uses an STDOUT logger by default. Uses Rails.logger when using Rails.

So that exceptions raised within streaming threads are clear and loud in logs, by default.

This default global callback applies to all dispatchers, but can be overridden globally with this in an initialiser

```ruby
Datastar.config.on_error do |error|
  # do something else here
end
```

```
RuntimeError (aaa):
/Users/ismasan/code/personal/experiments/dstar/app/controllers/search_controller.rb:70:in `block (2 levels) in search'
<internal:numeric>:237:in `times'
/Users/ismasan/code/personal/experiments/dstar/app/controllers/search_controller.rb:69:in `block in search'
/Users/ismasan/code/personal/gems/datastar/sdk/ruby/lib/datastar/dispatcher.rb:319:in `block (3 levels) in stream_many'
/Users/ismasan/.rbenv/versions/3.3.5/lib/ruby/gems/3.3.0/gems/activesupport-8.0.2/lib/active_support/execution_wrapper.rb:91:in `wrap'
/Users/ismasan/code/personal/gems/datastar/sdk/ruby/lib/datastar/rails_thread_executor.rb:8:in `block in spawn'
```

### DIFF
--- a/sdk/ruby/lib/datastar/configuration.rb
+++ b/sdk/ruby/lib/datastar/configuration.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'thread'
+require 'logger'
 
 module Datastar
   # The default executor based on Ruby threads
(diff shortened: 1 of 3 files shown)


bencroker (approved):

bencroker:
Merged, thanks!

Issue #768, opened 2025-03-16
Add type hints to the Python SDK


jmoppel:
### Feature Request

As a green field project that uses modern web tech, I think it makes sense to add rudimentary type hints to the Python SDK. It would be great to have this out of the box in a version 1.0 and would make the project appear polished and modern to future `datastar_py` users.

To be specific, I propose that anything user facing and/or improves user UX should be typed (assuming it doesn't damage readability too much). If folks agree, (off the top of my head) we could type hint objects like:

* Constants
* Function parameters and their output
* Classes (as needed)
* Public method parameters and their output
* Other stuff I've neglected off the top of my head

This list would likely not include variables inside functions/methods, or anything the user shouldn't be mucking about with. Adding type hints to internal stuff hurts readability too much for not enough benefit (IMO). Things like internal-only functions/classes and private methods are debatable and perhaps case-by-case.

If folks agree, I'm willing to help with this effort. Also, there would be the issue of which type checker to use. Finally, I would note that the generated code in `consts` would need to be changed by someone. I'm not sure how it is actually generated.

### Datastar Version

v1.0.0-beta.9

bencroker:
Paging @lllama.

bencroker:
Paging @gazpachoking.

gazpachoking:
I think we are pretty close on this, I'll take a look later today and fill in what's still missing.

PR #917, opened 2025-06-03
Python SDK: Allow specifying status code and headers with quark responses

gazpachoking:
- Allows specifying headers and status code with a quart datastar response
- Adds typing for quart's `make_datastar_response` refs #768
- Doesn't disable the default quart response timeout unless a generator is returned (only disables if we are streaming a response)


bencroker (approved):

bencroker:
Merged, thanks!

PR #918, opened 2025-06-04
Python SDK: Make response helpers opinionated about taking 0..N datastar events

### LINKED ISSUE #768: Add type hints to the Python SDK

gazpachoking:
A few main changes here:

## Opinionated Responses
Previously the response helpers were the thinnest possible wrapper on the native streaming response from the different frameworks. All of them have been redone to be much more specific about handling Datastar events, largely with typing, but also with functionality.
- `ServerSentEventGenerator` now generates `DatastarEvent`s (which are still just strings)
- All of the response handling now takes 0..N DatastarEvents (`None | DatastarEvent | Iterable[DatastarEvent] | AsyncIterable[DatastarEvent]
- If no datastar events are passed to a response, it is automatically turned into a 204
- A single datastar event can now be passed, without having to wrap it in a list or other iterable
- All of the frameworks response classes have been renamed to `DatastarResponse`.

Some examples, these work in all the frameworks except difference noted for sanic streaming.
```python
# 0 events, a 204
return DatastarResponse()
# 1 event
return DatastarResponse(ServerSentEventGenerator.merge_fragments("<div id='mydiv'></div>"))
# 2 events
return DatastarResponse([
    ServerSentEventGenerator.merge_fragments("<div id='mydiv'></div>"),
    ServerSentEventGenerator.merge_signals({"mysignal": "myval"}),
])
# N events, long lived stream (for all frameworks but sanic)
async def updates():
    while True:
        yield ServerSentEventGenerator.merge_fragments("<div id='mydiv'></div>")
        await asyncio.sleep(1)
return DatastarResponse(updates())
# for sanic
response = await datastar_respond(request)
# this is just a helper for the following now
# response = await request.respond(DatastarResponse())
while True:
    await response.send(ServerSentEventGenerator.merge_fragments("<div id='mydiv'></div>"))
    await asyncio.sleep(1)
```
## ServerSentEventGenerator
The event generator helper was previously including some options on every event, even if they were  set as the default. It now only includes the following options if the user specifies them:
- retry_duration, for all event types
- use_view_transitions, for merge and remove fragments
- only_if_missing, for merge signals
- auto_remove, for execute script

## Typing
I believe everything is now correctly typed. fixes #768 

## Future implications
Having a more specific type for DatastarEvent means that we could implement the implicit merge fragments from #809 more cleanly. If the response helper gets a DatastarEvent, it knows it's an event already, any other strings could then be assumed HTML, and auto wrapped in a merge-fragments.

## TODO
- [x] I used warnings.deprecated to mark the old response class names as deprecated but... that only arrived in python 3.13 we could:
  - [x] Just remove them. Not sure how many people are using the SDK, but current code would need to be updated or break.
  - Keep using deprecated decorator, and fall back to a passthrough on older pythons. This means it wouldn't emit the warning on older pythons.
  - Add a dependency on typing_extensions.
  - Roll our own dependency warning. Not quite as nice, because IDEs and type checkers detect the new decorator.
- [x] quart has `make_datastar_response` helper, which is slightly more idiomatic in that framework than a response subclass, but perhaps it makes sense to just deliver one anyway for parity?
  - [x] Switched to DatastarResponse so it works like the rest now.
- [x] sanic is the odd man out. Our more opinionated way doesn't map cleanly onto their normal streaming method. We could:
  - Not worry about it. (~~This is what I'll pick unless sanic people express an opinion.~~ this was a lie)
  - Add a simple helper for non-streaming datastar responses to make that case simpler.
  - Add an 'old style' response class which is more similar to the other frameworks
  - [x] Make a DatastarResponse which works like the other frameworks for simple (non-streaming) responses. It also sets the typing properly on response.send to indicate that it takes DatastarEvents.
- [x] Wait for attribute generator PR to be merged so we can update the readme without conflicts.

### DIFF
--- a/examples/python/django/ds/views.py
+++ b/examples/python/django/ds/views.py
@@ -3,7 +3,7 @@ import time
 from datetime import datetime
 
 from datastar_py.django import (
-    DatastarStreamingHttpResponse,
+    DatastarResponse,
     ServerSentEventGenerator,
     read_signals,
 )
@@ -65,7 +65,7 @@ async def updates_asgi(request):
             )
             await asyncio.sleep(1)
 
-    return DatastarStreamingHttpResponse(time_updates())
+    return DatastarResponse(time_updates())
 
 
 # WSGI Example
@@ -121,4 +121,4 @@ def updates_wsgi(request):
             )
             time.sleep(0.5)
 
-    return DatastarStreamingHttpResponse(time_updates())
+    return DatastarResponse(time_updates())
--- a/examples/python/fastapi/app.py
+++ b/examples/python/fastapi/app.py
@@ -13,7 +13,7 @@ from datetime import datetime
 
 import uvicorn
 from datastar_py.fastapi import (
-    DatastarStreamingResponse,
+    DatastarResponse,
     ReadSignals,
     ServerSentEventGenerator,
 )
@@ -80,7 +80,7 @@ async def time_updates():
 async def updates(signals: ReadSignals):
     # ReadSignals is a dependency that automatically loads the signals from the request
     print(signals)
-    return DatastarStreamingResponse(time_updates())
+    return DatastarResponse(time_updates())
 
 
 if __name__ == "__main__":
--- a/examples/python/fasthtml/advanced.py
+++ b/examples/python/fasthtml/advanced.py
@@ -17,7 +17,7 @@ from datetime import datetime
 from pathlib import Path
 
 import polars as pl
-from datastar_py.fasthtml import DatastarStreamingResponse, ServerSentEventGenerator
+from datastar_py.fasthtml import DatastarResponse, ServerSentEventGenerator
 from great_tables import GT
 from great_tables.data import reactions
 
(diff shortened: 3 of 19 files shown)


lllama:
I'd just remove them as we're pre1.0.

bencroker:
Agreed, no need to deprecate things pre v1.

gazpachoking:
I figured a nice way to do the sanic stuff as well now. My only question is if we should just go ahead and make a quart DatastarResponse rather than the make_datastar_response helper. Then all of them would be used pretty much the same (except sanic streaming.)

gazpachoking:
Moved quart to the DatastarResponse style. Parity across all frameworks except for sanic streaming.

bencroker (approved):

bencroker:
Merged 🚀

PR #922, opened 2025-06-07
Python SDK: Allow strings and dicts to become implicit merge fragments/signals

gazpachoking:
Continuing on an idea from #809

This allows sending merge-fragments and merge-signals events with the default options by just passing a string or dict into the `DatastarResponse`. This would not replace the current way, just be a shortcut for the most common two events.

e.g.
```python
return DatastarResponse(ServerSentEventGenerator.merge_fragments("<div id='mydiv'></div>"))
# can be shortened to
return DatastarResponse("<div id='mydiv'></div>")

# similarly you can provide a dict to do a merge-signals
return DatastarResponse([
    "<div id='mydiv'></div>",
    {"signala": "signalval"}
])

# Also works in long lived generators
async def updates():
    yield "<div id='mydiv'></div>"
    yield {"signala": "signalval"}
return DatastarResponse(updates())

# And in sanic style streaming
response = await datastar_respond(request)
await response.send("<div id='mydiv'></div>")
await response.send({"signala": "signalval"})

# htpy elements or fasttags can be passed in as well and become merge-fragments
return DatastarResponse(div("#mydiv"))
return DatastarResponse(Div(id="mydiv"))

# If you need non-default options for the merge you still do it the normal way.
return DatastarResponse([
    ServerSentEventGenerator.merge_fragments("<div></div>", merge_mode=FragmentMergeMode.APPEND, selector="#whatever"),
    # both styles can be mixed
    {"signal": "val"},
    # Other events are still done normally
    ServerSentEventGenerator.execute_script("console.log('foo')"),
])
```

I'd like some feedback on this one, not sure if it's too much magic, since it really isn't hard to call the SSE generator.

### DIFF
--- a/examples/python/quart/app.py
+++ b/examples/python/quart/app.py
@@ -12,7 +12,6 @@ from datetime import datetime
 
 from datastar_py.quart import (
     DatastarResponse,
-    ServerSentEventGenerator,
     read_signals,
 )
 
@@ -63,13 +62,17 @@ async def updates():
 
     async def time_updates():
         while True:
-            yield ServerSentEventGenerator.merge_fragments(
-                f"""<span id="currentTime">{datetime.now().isoformat()}"""
-            )
+            # DatastarResponse automatically turns strings into a merge_fragments event
+            yield f"""<span id="currentTime">{datetime.now().isoformat()}</span>"""
+            # yield ServerSentEventGenerator.merge_fragments(
+            #     f"""<span id="currentTime">{datetime.now().isoformat()}"""
+            # )
             await asyncio.sleep(1)
-            yield ServerSentEventGenerator.merge_signals(
-                {"currentTime": f"{datetime.now().isoformat()}"}
-            )
+            # dicts are automatically turned into a merge_signals event
+            yield {"currentTime": f"{datetime.now().isoformat()}"}
+            # yield ServerSentEventGenerator.merge_signals(
+            #     {"currentTime": f"{datetime.now().isoformat()}"}
+            # )
             await asyncio.sleep(1)
 
     return DatastarResponse(time_updates())
--- a/examples/python/sanic/app.py
+++ b/examples/python/sanic/app.py
@@ -115,11 +115,8 @@ async def updates(request):
             )
         )
         await asyncio.sleep(1)
-        await response.send(
-            ServerSentEventGenerator.merge_signals(
-                {"currentTime": f"{datetime.now().isoformat()}"}
-            )
-        )
+        # a dict response becomes a merge-signals event implicitly
+        await response.send({"currentTime": f"{datetime.now().isoformat()}"})
         await asyncio.sleep(1)
 
 
--- a/sdk/python/README.md
+++ b/sdk/python/README.md
@@ -1,3 +1,5 @@
+from datastar_py import ServerSentEventGeneratorfrom datastar_py import ServerSentEventGenerator
+
 # datastar-py
 
 The `datastar-py` package provides backend helpers for the [Datastar](https://data-star.dev) JS library.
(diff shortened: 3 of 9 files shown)


cro:
No, not too much magic, I wrote helper functions that basically do the same thing for my current project. This would be a good addition.

bencroker:
I’m not sure whether adding magic is a good idea here. You’re promoting a lazy style of coding which doesn’t work beyond two event types with default values. I’d like to hear from more users of the Python SDK before making a decision.

dalito:
In the current version, I do not find it elegant that one has to go back to `ServerSentEventGenerator.merge_fragments("<div></div>", merge_mode=FragmentMergeMode.APPEND,...)` if options are passed.
I also wonder if it would be cleaner to use [functools.singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) to dispatch based on type instead of the multiple "if isinstance()" checks.

Overall, I like the idea of such a simplified interface.

axelknock:
Since Datastar v1 will support implicit merge-fragments/signals/execute responses from non-EventStream responses, I think that each of the framework's respective HTML/JSON/text response types is suitable for all of the use-cases this would cover.

gazpachoking:
> I’m not sure whether adding magic is a good idea here. You’re promoting a lazy style of coding which doesn’t work beyond two event types with default values. I’d like to hear from more users of the Python SDK before making a decision.

Yep, I had the same reservation. My intuition says those two event types with default values are the most common needs, so it seems nice to have a shortcut. On the other hand, I agree in that the magic can obscure what's actually happening to new users or readers of the code.



> In the current version, I do not find it elegant that one has to go back to `ServerSentEventGenerator.merge_fragments("<div></div>", merge_mode=FragmentMergeMode.APPEND,...)` if options are passed.

I don't think there is good solution to that. My solution would be to not merge this PR, so the interface is always the same of calling the events explicitly every time.

> Since Datastar v1 will support implicit merge-fragments/signals/execute responses from non-EventStream responses, I think that each of the framework's respective HTML/JSON/text response types is suitable for all of the use-cases this would cover.

Agreed, that does cover some cases, although I like the explicitness of returning a DatastarResponse for every response to datastar, and letting html responses be for the browser (I know, ironic that I'm arguing for explicit being better in a PR about making other things implicit 😅 ). The bigger advantage to this way is that it remains an SSE, so you can deliver more than one event still. I also had (have) similar reservations about datastar allowing other content types as I do to this PR.

Ultimately I could still be swayed either direction. Everyone here has good points, and I don't know which direction the scales should tip.

bencroker:
I wouldn’t recommend any other SDK do implement this, so I don’t think the Python one should either. 

If you yourself are unconvinced, I’d say we shelf this.

delaneyj:
I have no opinion on this other than @gazpachoking seems to have the pulse and should be the final arbiter but will be forced to support it and never leave.

gazpachoking:
I am still slightly on the in favor side, but I think shelving it for now and waiting to see how I/others feel later is a pretty reasonable plan.

1363V4:
Also keep in mind the IDE autocomplete.
I type `SSE.` and see the available functions, `SSE.merge_fragments()` and I see the parameters...
Pretty convenient and we would lose that.

gazpachoking:
> Also keep in mind the IDE autocomplete.
> I type `SSE.` and see the available functions, `SSE.merge_fragments()` and I see the parameters...
> Pretty convenient and we would lose that.

This would not take away or change anything about the current method or ServerSentEventGenerator. All current code would work exactly the same. It would merely allow not using it (but you still could use it) for merge_fragments and merge_signals if you are using the default options.

1363V4:
After giving it some thought, I think it is too much magic.
I like the concept separation of:
> DatastarResponse sets the relevant headers (for the browser)
> SSE.xxx is just a bunch of data for datastar to use

With a syntax like `DatastarResponse("<div id='mydiv'></div>")`, people might get under the impression that we only send HTML, and then "magic happens".

bencroker:
Closing.

PR #923, opened 2025-06-07
Python SDK: Move versioning to pyproject.toml

gazpachoking:
This moves the version to pyproject.toml and bumps it up to 0.5.0. The version of the pypi release had gotten out of sync from the version in `__about__`. Also added myself to authors and removed the bumpversion config as we could use `uv version --bump major|minor|patch` to update the version in the pyproject.

@lllama Should make the decision on this one, as he's doing the pypi releases and I don't know how this would affect that current workflow.

### DIFF
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -1,14 +1,15 @@
 [project]
 name = "datastar-py"
 description = "Helper functions and classes for the Datastar library (https://data-star.dev/)"
+version = "0.5.0"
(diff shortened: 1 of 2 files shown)


gazpachoking:
image

lllama:
LGTM. I'm using `uv` to publish to pypi anyway. Didn't know it did version bumping 🤣

bencroker (approved):

bencroker:
Merged, thanks!

PR #925, opened 2025-06-08
Python SDK: Add a decorator to auto-wrap responses in DatastarResponse

gazpachoking:
This is the second idea split out from #809

The main idea is to simplify streaming response handlers so that a second function isn't required inside your route function.

```python
from datastar_py import ServerSentEventGenerator as SSE

# The current way
async def some_route():
    async def stream_updates():
        for event in wait_for_events():
            yield SSE.merge_fragments("<div id='aoeu'>Updates from the event.</div>")
    return DatastarResponse(stream_updates())

# With this decorator
@datastar_response
async def some_route():
    for event in wait_for_events():
        yield SSE.merge_fragments("<div id='aoeu'>Updates from the event.</div>")
```

It isn't limited to streaming responses though. It works on any sync or async function or generator function that returns/yields 0..N datastar events.

```python
@datastar_response
async def some_route():
    if something:
        return # A 204
    else:
        return SSE.merge_signals({'my': 'signals'})

@datastar_response
def some_route():
    return SSE.merge_fragments("<div id='aoeu'>Some fragment.</div>")

@datastar_response
async def some_route():
    return (
        SSE.merge_fragments("<div id='aoeu'>Some fragment.</div>"),
        SSE.merge_signals({"and": "signals"}),
        SSE.execute_script("console.log('and script');")
    )

@datastar_response
async def some_route():
    return some_generator()
```

Works with all frameworks with the same syntax. ~~but sanic streaming is different, so it doesn't help there.~~ (You don't need a separate generator function in sanic like you do for the others, but the @datastar_response syntax works the same across all frameworks.)

TODO:
- [ ] Update examples/readme
  - How to update examples? Want to show both the long ways and the shortcuts. 
- [ ] Does it need to be able to set headers? It _could_ be changed to allow that like `@datastar_response(headers={})` I'm leaning towards leaving that out though.


gazpachoking:
Okay, switched this to be on top of latest changes and added info to the readme. (also fixed up a few things in the readme) This is good to go.

bencroker:
Thanks.

PR #926, opened 2025-06-09
Python SDK: Allow overriding default response headers

gazpachoking:
When using `DatastarResponse` headers passed in get merged with the default headers, so there was no way to remove headers included in the default you didn't want (e.g. the x-accel-buffering for nginx.) This also allows sending an extra header with all datastar responses. User provided defaults can be set on the `DatastarResponse` classes with this change.

```python
DatastarResponse.default_headers = {**SSE_HEADERS, "x-my-extra-header": "foo"}

# ...

    return DatastarResponse()  # user's default headers used
    return DatastarResponse(headers={"x-another": "header"})  # Another header merged in with the defaults
```

### DIFF
--- a/sdk/python/src/datastar_py/django.py
+++ b/sdk/python/src/datastar_py/django.py
@@ -23,6 +23,8 @@ __all__ = [
 class DatastarResponse(_StreamingHttpResponse):
     """Respond with 0..N `DatastarEvent`s"""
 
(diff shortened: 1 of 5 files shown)


bencroker (approved):

bencroker:
Merged, thanks!

PR #938, opened 2025-06-15
sdk/dotnet csharp

SpiralOSS:
C# version of the SDK.

@bencroker The last two commits are probably what you are interested in. They are updates to the build for Consts and HelloWorld, and the nuget workflows. I split this so there's an fsharp and a csharp nuget that will need to get made if this gets approved.


bencroker (approved):

bencroker:
Looks great! Merged, thanks.

PR #940, opened 2025-06-18
Update CONTRIBUTING.md - Duplicated SDKs block?

eduwass:
I noticed that this block is duplicated on the page. Could it be a mistake?


bencroker:
PRs must be made against the `develop` branch, in which there is no duplication.

eduwass:
You are totally right lol, my bad @bencroker I don't know how I ended up there...

PR #941, opened 2025-06-18
Add some more PatchElement examples to the SDK ADR

gazpachoking:
Add some more examples to clarify PatchElements events in the ADR. Particularly showing the remove mode with and without elements specified.

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -144,6 +144,47 @@ data: elements </div>
 ```
 </details>
 
+<details>
+<summary>Patch elements based on their id</summary>
+
+```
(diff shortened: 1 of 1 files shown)


bencroker:
Thanks for the PR! I’m not fond of the use of `: ` for comments, as it could cause confusion. Can you please split the `details` element into one per example, named accordingly?

gazpachoking:
Okay, split out all the new examples, and put them up with the existing examples. Also explicitly documented the `elements` argument now.

bencroker (approved):

bencroker:
Merged, thanks!

PR #952, opened 2025-06-26
Refactor Rust SDK for RC

### LINKED ISSUE #909: refactor rust SDK to make use of frameworks their SSE support
- axum and rocket already have SSE support
- for rama I am adding it (https://github.com/plabayo/rama/issues/570)

We can do this as part of a 0.3 release of datastar rust crate together with other refactors and improvements (eg https://github.com/starfederation/datastar/issues/906)

I’ll pick this up myself somewhere in next weeks

GlenDC:
- frontend examples broken for now, due to missing datastar.js bundle build (SDK maintainers will be informed when available)
- rely on web frameworks (Rocket, Axum) for actual SSE code, and have datastar focus only on datastar Closes #909
- adapt examples and cleanup code + deps
- remove rama as built-in framework to support (do reference to the rama datastar module, which is to be used as the datastar rust sdk for any rama user)


bencroker:
While not required, all SDKs are continuing to implement `ExecuteScript`. Can you please add it back in, following the convention used by the Go SDK?

GlenDC:
> While not required, all SDKs are continuing to implement `ExecuteScript`. Can you please add it back in, following the convention used by the Go SDK?

I do not agree. I do not see it any different then other elements. Where does sugar stop at that point. Plus once you need anything with scripts that is not provided by the sugar you are once again at patch elements anyway...

bencroker:
> I do not agree.

With what exactly?

> I do not see it any different then other elements.

Script tags need special handling, so we do some magic to ensure that browsers execute them. The `ExecuteScript` function means that users can execute JS code without having to think in terms of a `script` tag. Yes, it’s just a convenience at this point, but one that all SDKs should implement so that it makes moving between SDKs easier, including the code snippets we provide in the docs.

GlenDC:
> Script tags need special handling, so we do some magic to ensure that browsers execute them. The `ExecuteScript` function means that users can execute JS code without having to think in terms of a `script` tag. Yes, it’s just a convenience at this point, but one that all SDKs should implement so that it makes moving between SDKs easier, including the code snippets we provide in the docs.

Fair enough. I checked the Go SDK. I agree now. I'll revive it :)
(tomorrow)

bencroker:
Please rebase to `release-candidate`.

GlenDC:
@bencroker ready for review / merge.
Updated to RC13, also updated deps + fix newest lints for latest rust versions.
Rama was removed as well, and sugar Execute Script was preserved.

CLI Fails due to tailwind download

bencroker:
Merged, thanks! I also ran a build and committed in https://github.com/starfederation/datastar/commit/8b2e835a84fc58674cd731450304e55d2acd9103.

PR #955, opened 2025-07-01
Python SDK: Tweak execute_script sugar method

gazpachoking:
Updates the Python SDK execute_script sugar:
- changes `auto_remove` to default to True (aligns with go sdk default)
- Uses `onload` instead of `data-on-load` for auto_remove so as not to depend on data-on-load plugin


bencroker:
Merged, thanks!

PR #959, opened 2025-07-01
Ruby fix execute script, and fix constants generation

ismasan:
In this PR:

* Update Ruby constants template, regenerate constants and fix code.
* Fix `#execute_script` to use `data-effect="el.remove()"` instead of `onload="this.remove()"`
* Document build task so that I don't forget about it next time.


bencroker:
Looking good! One thing to change is that the attribute default should be empty.

ismasan:
@bencroker I've removed that attribute default check since it's not needed anymore.

bencroker:
Thanks!

PR #961, opened 2025-07-03
Clojure SDK: ADR changes and 2 new features

JeremS:
This PR mainly contains the changes for the new ADR:

- the public API functions have been renamed
- the remove-elements and the execute-script functionality is now sugar on top of patch-elements
- the hello world example and all the code snippets for the website are updated to the new names
- the SDK examples have also been updated to the new API

I used these changes as an opportunity to review some things:

- I moved the clj-kondo config to allows other lib to use the starfederation.datastar.clojure ns for their clj-kondo configs.
- I refactored the unit tests for the core API, they now should be easier to understand, maintain and are more repl friendly

I also added 2 new features:
- support for brotli compression as a library see the `sdk-brotli` directory (and the associated commit)
- an alternative http-kit api that works better with the ring model at the cost of using 1 middleware (see the http-kit2 namespace)

All in all, the breaking changes are ADR related, the new features are additive and optional.

Cheers!


bennyandresen:
Thanks a lot @JeremS!

For information: After updating to the newest SDK from this PR I merely had to change `merge-fragments!` to `patch-elements!` in my application code and everything continued to work.

Went from rc11 to rc13.

Brotli is currently not working for me, but we'll investigate this via Discord.

bencroker:
Is this ready for merging?

JeremS:
Brotli on Http-kit works. When using jetty that's another story. I suspect it's a Jetty buffering issue not a SDK one. I will add a notice to the Brotli lib's README to that effect. 

@bencroker Usually @andersmurphy & @Ramblurr review the Clojure SDK. I think it's fine to be merged but I'd also like to get their go ahead. I should have mentioned them in this PR sooner.

anastygnome:
Jetty will not work  if you don't set the settings as per 

https://github.com/jetty/jetty.project/issues/4317

JeremS:
@anastygnome Hi, I am not sure what you mean, what settings should we set? For info the SDK's GZIP write profile works with Jetty, it's the Brotli one that does not.

Ramblurr:
I am looking at this today.

anastygnome:
> @anastygnome Hi, I am not sure what you mean, what settings should we set? For info the SDK's GZIP write profile works with Jetty, it's the Brotli one that does not.

I'll check this when I get to that point with the jetty buffer, but this is the same kind of buffer issue we encounter with gzip (even though it works)

sdk/clojure/sdk-brotli/deps.edn
```diff
@@ -0,0 +1,3 @@
+{:paths ["src/main"]
```
Ramblurr, line 1: Do we want to add build.clj and a build script to produce a clojars artifact for this new brotli library?

sdk/clojure/src/dev/examples/snippets.clj
```diff
@@ -10,10 +10,10 @@
...
 
 ;; multiple_events
-(d*/merge-fragment! sse "<div id=\"question\">...</div>")
-(d*/merge-fragment! sse "<div id=\"instructions\">...</div>")
-(d*/merge-signals! sse "{answer: '...'}")
-(d*/merge-signals! sse "{prize: '...'}")
+(d*/patch-elements! sse "<div id=\"question\">...</div>")
+(d*/patch-elements! sse "<div id=\"instructions\">...</div>")
+(d*/patch-elements! sse "{answer: '...'}")
+(d*/patch-elements! sse "{prize: '...'}")
```
Ramblurr: You changed these from `merge-signals!` -> `patch-elements!`. Shouldn't they be `patch-signals!`?

sdk/clojure/src/dev/examples/snippets.clj
```diff
@@ -27,10 +27,10 @@
   (->sse-response request
     {on-open
      (fn [sse]
-       (d*/merge-fragment! sse
+       (d*/patch-elements! sse
          "<div id=\"question\">What do you put in a toaster?</div>")
 
-       (d*/merge-signals! sse "{response: '', answer: 'bread'}"))}))
+       (d*/patch-elements! sse "{response: '', answer: 'bread'}"))}))
```
Ramblurr: Shouldn't this be `patch-signals!`?

bencroker:
Can we get this finalised today for the public RC?

sdk/clojure/sdk-brotli/deps.edn
Ramblurr: this is resolved in the next PR @JeremS is working on in the v1-prep branch https://github.com/JeremS/datastar/tree/v1-prep

Ramblurr (approved):
@bencroker My comments only affect the testing/example files, not the core SDK code.  **IMO it can be merged as is.** I've tested that the core functionality works in my apps (though I haven't used the new http-kit2 ns yet).

@JeremS I'm approving these changes despite the comments above, maybe you can address them later?

bencroker:
Let’s get these resolved first.

sdk/clojure/src/dev/examples/snippets.clj
JeremS: Good catch

sdk/clojure/src/dev/examples/snippets.clj
JeremS: Also good catch

JeremS:
I have fixed the mistakes @Ramblurr  caught. They were in example snippets I use to test not the ones that go into the D* website though. I checked, these mistake are only in my tests, not on the website.

@Ramblurr Thanks for the review.
@bencroker  We are good to go!

PR #963, opened 2025-07-04
Python SDK: Update attribute generator for new RC

gazpachoking:
- Adds new attributes and modifiers
- The data-on-X attributes provided by datastar now have their own attributes instead of just being lumped into data-on (e.g. `data.on('load', 'expr')` -> `data.on_load('expr')`) The old way still works as well.
- Marks which attributes are PRO features
- Refactors a bit to make things simpler

Still WIP until new site comes out and I can verify everything against the docs.


bencroker:
Please proceed and let me know when I can merge.

PR #965, opened 2025-07-06
Updated the SDK test suite to the new ADR

mailq:
Requesting for comments and review as I am not sure if I understood the new ADR correctly.


sdk/test/get-cases/executeScriptWithDefaults/output.txt
```diff
@@ -1,4 +1,4 @@
-event: datastar-execute-script
-data: script console.log('hello');
+event: datastar-patch-elements
+data: elements <script type="module" data-effect="el.remove()">console.log('hello');</script>
```
bencroker: `type="module` should be removed.

mailq: This is where I am unsure. If you explicitly request that in the input, it should be present in the output. The ADR does not state, that the explicit attribute "type" should be filtered out.

bencroker: It should not be in the input either, since it is not a default (it used to be).

bencroker:
Merged, thanks!

PR #968, opened 2025-07-07
Added test for multi line signals

mailq:
this is different from "multi-lined" signals.

The former allows signals spanning multiple lines in the SSE event.
The latter allows signal values to contain line-feeds.

Also the test for the latter variant had the wrong escape sequence for line-feeds in them.


bencroker:
Merged, thanks!

PR #971, opened 2025-07-08
Python SDK: Add remove elements sugar

gazpachoking:
This improves the ergonomics for removing elements:
- Adds `remove_elements` sugar which takes a selector
- Allows calling `patch_elements` without elements when remove mode is used


bencroker:
Thanks!

PR #974, opened 2025-07-08
Fixes for sdk tests

gazpachoking:
- Adds mode/selector to the executeScript tests output
- Delete extra newlines in sendTwoEvents. SSE spec says two newlines to delimit events, not 3.
- Makes sure test-post.sh is compatible with versions of curl that don't support --json
- Change the test for adding attributes in executeScript to only use string attributes


GlenDC (approved):
LGTM, I think this might fix all remaining diffs I had open for the SDKs I maintain

bencroker:
Lovely, thanks!

PR #975, opened 2025-07-08
Python SDK: Add a test server to run the sdk tests against

gazpachoking:
This adds a server to run the language agnostic sdk tests against for python.

Also tweaks a few things for full compatibility in the tests:
- Allows an already formatted json string for patch_signals
- Allows a dictionary for the `attributes` argument of execute_script
- Doesn't send the mode of 'outer' for patch_elements, as it's default
- Puts the event type first in the sse events


bencroker:
Lovely, thanks!

PR #976, opened 2025-07-08
Ruby release candidate fixes

ismasan:
Minor improvements to support the new release candidate:

* Explicitly require `logger` and `json`, which in future Ruby versions won't be part of the standard library.
* Make `#execute_script()` optionally accept a JSON-encoded string as signals argument, instead of a Hash.

```ruby
sse.execute_script <<~JSON
 {
   "name": "Joe",
   "age": 34
 }
JSON
```

Multiline JSON strings will be sent as multiple `data: signals` lines, as per the ADR spec.


bencroker:
Lovely, thanks!

PR #977, opened 2025-07-08
Patch/rust sdk test suite compliance

GlenDC:


GlenDC:
Also opened a PR in axum as the only remaining diffs are because of an odd thing in axum itself: https://github.com/tokio-rs/axum/pull/3403

in meanwhile this PR can already be merged

PR #978, opened 2025-07-09
Ruby fix event end and pass bash tests

ismasan:
* Fix SSE event termination (single `\n` instead of double)
* Fix test app and make all (available) bash tests pass
* Fix top-level Ruby example to use release-candidate JS bundle


bencroker:
Thanks!

PR #980, opened 2025-07-10
Update cdn version for python examples

gazpachoking:
Updates the python examples so that they get datastar from the CDN at the version of datastar that the sdk was built for.

Opening as draft for now because I'm not sure this is the right approach.

Another possibility is to change all of them to serve datastar themselves as static files from the bundles folder in the repo.

### DIFF
--- a/examples/python/django/ds/views.py
+++ b/examples/python/django/ds/views.py
@@ -2,6 +2,7 @@ import asyncio
 import time
 from datetime import datetime
 
+from datastar_py.consts import VERSION
 from datastar_py.django import (
     DatastarResponse,
     ServerSentEventGenerator,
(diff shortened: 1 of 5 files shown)


jmoppel:
Ideally, the CDN would have a `stable` URL that redirects to the latest version. If that is an option once v1 is released, then we would just use the stable link, and suffer through hard coding a known compatible version for now. Either way, this is likely temporary growing pains, considering Datastar is targeting a long-term stable API and the examples are basic enough they're very unlikely to break.

gazpachoking:
I guess `@main` would be the stable version. I think it wasn't getting updated with the betas, which is why we we had to update it multiple times to keep things working. I guess that's a better idea than this. 👍

PR #982, opened 2025-07-10
last minute 1.0 rc1 bugfixes for rust sdk

GlenDC:
ready to ship 0.3 of the crate


GlenDC:
@bencroker please publish to 0.3 and announce in #general-rust when you did, with this PR merged. should be ok.
Only sdk test diff we now still have is already fixed in https://github.com/tokio-rs/axum/commit/617594cd4a3e2b90f2d7c061fecfbbbfea713b3d, so in a future dep update that will fix itself

bencroker (approved):

PR #984, opened 2025-07-10
Updated TypeScript SDK 

eduwass:
This PR introduces updated TypeScript SDK for Datastar with support for Node.js, Deno, and Bun runtimes.

## Features

- **Multi-runtime support**: Node.js, Deno, and Bun compatibility
- **Spec-compliant**: Implements the new SDK specification
- **Abstract architecture**: Extensible design for future runtime implementations
- **Improved documentation and testing**: Rewrote readme, examples and added some improvements for testing workflow

## Testing

Seems like all tests passing: 20/20 test cases 

Used the language agnostic tests from [sdk/test](https://github.com/starfederation/datastar/tree/develop/sdk/test) thanks to #974 !

[How to run tests](https://github.com/eduwass/datastar/blob/feature/typescript-sdk/sdk/typescript/README.md#testing)


bencroker:
Could you please give this a look and review, @jmstevers?

sdk/typescript/README.md
```diff
@@ -1,96 +1,279 @@
...
-     stream.mergeFragments(`<div id="toMerge">Hello <span data-text="$foo">${reader.signals.foo}</span></div>`);
+    // Patch signals
+    stream.PatchSignals(JSON.stringify({ foo: reader.signals.foo }));
+    
+    // Patch DOM elements
+    stream.PatchElements(`
+        <div id="toMerge">
+            Hello <span data-text="$foo">${reader.signals.foo}</span>
+        </div>
+    `, { mode: "outer" });
```
ghost: Mode `outer` is redundant here because its the default

sdk/typescript/src/consts.ts
```diff
@@ -2,7 +2,7 @@
 
 export const DATASTAR = "datastar" as const;
 export const DATASTAR_REQUEST = "Datastar-Request";
-export const VERSION = "1.0.0-RC.1";
+export const VERSION = "1.0.0-RC.12";
```
ghost, line 5: This file should be regenerated. Don't manually modify the consts.ts file because it gets auto-generated

sdk/typescript/package.json
```diff
@@ -1,11 +1,16 @@
 {
   "name": "@starfederation/datastar-sdk",
-  "version": "1.0.0-beta.11",
+  "version": "1.0.0-RC.12",
```
ghost:
```suggestion
  "version": "1.0.0-RC.1",
```

sdk/typescript/examples/node/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "1.0.0",
+  "main": "node.js",
+  "scripts": {
+    "start": "node node.js"
+  },
+  "author": "",
+  "license": "ISC",
+  "description": "",
+  "dependencies": {
+    "datastar-sdk": "^1.0.0-beta.11"
```
ghost:
```suggestion
    "datastar-sdk": "^1.0.0-RC.1"
```

sdk/typescript/examples/bun/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "type": "module",
+  "private": true,
+  "devDependencies": {
+    "@types/bun": "latest"
+  },
+  "peerDependencies": {
+    "typescript": "^5"
+  },
+  "dependencies": {
+    "datastar-sdk": "^1.0.0-beta.11"
```
ghost:
```suggestion
    "datastar-sdk": "^1.0.0-RC.1"
```

sdk/typescript/package.json
```diff
@@ -18,12 +23,20 @@
...
+    "npm/script/package.json",
+    "npm/esm/src/",
+    "npm/esm/package.json",
+    "npm/package.json",
+    "npm/README.md",
+    "README.md"
+  ],
   "dependencies": {
     "deepmerge-ts": "^7.1.4"
   },
```
ghost:
```suggestion
```
Unused dependency

sdk/typescript/src/abstractServerSentEventGenerator.ts
```diff
@@ -2,29 +2,67 @@ import {
...
+   * @param jsonString - The JSON string to validate
+   * @throws {Error} If the JSON is invalid
+   */
+  private validateJSON(jsonString: string): void {
+    try {
+      JSON.parse(jsonString);
+    } catch (error) {
+      throw new Error(`Invalid JSON provided: ${error instanceof Error ? error.message : 'Unknown error'}`);
+    }
+  }
```
ghost:
```suggestion
```
PatchSignals string is not required to be a valid JSON value

sdk/typescript/src/abstractServerSentEventGenerator.ts
```diff
@@ -80,130 +118,87 @@ export abstract class ServerSentEventGenerator {
...
   ): ReturnType<typeof this.send> {
-    const eventOptions = options || {} as DatastarEventOptions;
-    const pathsArray = typeof paths === "string"
-      ? paths.split(" ")
-      : paths.flatMap((path) => path.split(" "));
+    // Validate required parameters
+    this.validateRequired(signals, 'signals');
+    
+    // Validate JSON format
+    this.validateJSON(signals);
```
ghost:
```suggestion
```
PatchSignals string is not required to be a valid JSON value

ghost (changes requested):

sdk/typescript/README.md
eduwass: Updated!

sdk/typescript/package.json
eduwass: Hm.. looks like this is actually used [here](https://github.com/eduwass/datastar/blob/56eaecc2a10341666ba16b94a4cb5dc6a9a4c0a6/sdk/typescript/src/web/serverSentEventGenerator.ts#L4) tho.

I've kept it because it was part of the previous implementation but we could probs remove... Having a look.

eduwass: OK all good to remove! https://github.com/starfederation/datastar/pull/984/commits/59d895b16da8d8757e58e725834a695724e24faf

eduwass:
Thanks for having a look @jmstevers !

I've just applied a set of edits based on your feedback!

eduwass:
BTW I was going through the new site (awesome job btw 💯) and I noticed minor inconsistency in this TS code snippet here:

002482@2x

https://data-star.dev/guide/backend_requests_sse_events#sse-events

Notice the different casing `patchElements` and `PatchSignals` which one should it be? Right now implementation uses PascalCase but happy to change to what you guys prefer, I think in TS camelCase is more common.

bencroker:
You tell me and I’ll fix. Please check all TS code snippets on the site (including in the how-tos), and provide any missing snippets, if you could.

Regaez:
> Notice the different casing `patchElements` and `PatchSignals` which one should it be? Right now implementation uses PascalCase but happy to change to what you guys prefer, I think in TS camelCase is more common.

I would vote for camel case, as it is the convention in JS/TS for methods.

eduwass:
OK updated @bencroker !

Having a look at the code snippets for the site now.

eduwass:
OK just skimmed through the site @bencroker @delaneyj 

- https://data-star.dev/guide/backend_requests_sse_events#sse-events
002483@2x

- https://data-star.dev/guide/backend_requests_sse_events#backend-actions

002484@2x


AFAIK it's only these 2 places where casing should be fixed: `stream.PatchSignals` → `stream.patchSignals`

sdk/typescript/src/node/serverSentEventGenerator.ts
```diff
@@ -100,7 +101,7 @@ export class ServerSentEventGenerator extends AbstractSSEGenerator {
    *
    * @returns An object containing a success boolean and either the client's signals or an error message.
    */
-  static async readSignals(request: IncomingMessage): Promise<
+  static async ReadSignals(request: IncomingMessage): Promise<
```
Regaez: As this is a method of `ServerSentEventGenerator`, should it also be camel case?

eduwass: Yes @Regaez absolutely, good catch!

Fixing...

sdk/typescript/examples/node/node.js
```diff
@@ -10,7 +9,7 @@ const server = createServer(async (req, res) => {
     const headers = new Headers({ "Content-Type": "text/html" });
     res.setHeaders(headers);
     res.end(
-      `<html><head><script type="module" src="https://cdn.jsdelivr.net/gh/starfederation/datastar@v1.0.0-beta.1/bundles/datastar.js"></script></head><body><div id="toMerge" data-signals-foo="'World'" data-on-load="@get('/merge')">Hello</div></body></html>`,
+      `<html><head><script type="module" src="https://cdn.jsdelivr.net/gh/starfederation/datastar@release-candidate/bundles/datastar.js"></script></head><body><div id="toMerge" data-signals-foo="'World'" data-on-load="@get('/merge')">Hello</div></body></html>`,
```
Regaez: Should these examples still be tracking the release candidate branch for the datastar bundle? Wouldn't it be more reliable to target a tagged release in the CDN (now that it's gone public)?

sdk/typescript/src/consts.ts
```diff
@@ -14,53 +14,48 @@ export const DefaultSseRetryDurationMs = 1000;
...
-
-
 // #region Default booleans
 
 // Should elements be patched using the ViewTransition API?
 export const DefaultElementsUseViewTransitions = false;
 
 // Should a given set of signals patch if they are missing?
-export const DefaultPatchSignalsOnlyIfMissing = false;
+export const DefaultpatchSignalsOnlyIfMissing = false;
```
Regaez: I think a find/replace for `patchSignals` might have accidentally changed this:
```suggestion
export const DefaultPatchSignalsOnlyIfMissing = false;
```

sdk/typescript/examples/node/node.js
eduwass: Yes @Regaez, thanks for your eyes on this, again great point!

The official site uses the URL:
```
https://cdn.jsdelivr.net/gh/starfederation/datastar@main/bundles/datastar.js
```
Not sure what the best approach is — should we stick with main, or use a tagged version instead?

Since this CDN URLs appear both in:
- test files
- examples (which users typically copy-paste)

... maybe it makes sense for examples to track main, while tests should be tied to a specific version?

sdk/typescript/src/consts.ts
eduwass: Yes!! Thanks for catching this, fixed!

sdk/typescript/examples/node/node.js
rmcsharry: I agree with @eduwass ' suggestion - I was testing moving from beta to rc.1 and I copy/pasted from the example, and it took me a while to realise I had miss-matched versions.

Regaez: I guess it might depend on how stable main will remain? Probably fine to do that for now and maybe switch to use something like the `@latest` tag post-release? (I believe it isn't updated yet because of the RC versioning)

eduwass: OK folks, I think I'm going to do this:

- Tests use a fixed version (current):
```
https://cdn.jsdelivr.net/gh/starfederation/datastar@v1.0.0-RC.1/bundles/datastar.js
```
This ensures tests are tied to a specific release and remain stable. When a new version is out, we need to tests stuff again and update this.

- Examples will use main:
```
https://cdn.jsdelivr.net/gh/starfederation/datastar@main/bundles/datastar.js
```
Since these are meant to be copy-pasted by users, it’s helpful to keep them up-to-date with the latest API. Which is what is shown in the Getting Started documentation as well, I think this avoids confusion for new users as well as not having to maintain so often.

Hopefully that makes sense!

eduwass: Fixed in: https://github.com/starfederation/datastar/pull/984/commits/8b76bfffb99dcafd5e14052e8554139c4167de80

sdk/typescript/src/consts.ts
rmcsharry, line 5: @eduwass Everywhere else in the PR it is `-RC.1`.

So shouldn't this also say `-RC.1` and not `-RC.12`?

eduwass: Ah yes, @jmstevers pointed out this file is auto-generated see: https://github.com/starfederation/datastar/pull/984#discussion_r2199810725

So I understand we don't have to change this.

Regaez (commented):
Great work @eduwass! No further comments from me. 

For what it's worth, as a non-maintainer, looks good to me. 👍

eduwass:
Thanks @Regaez ! :)

sdk/typescript/examples/node/node.js
bencroker: Please use `https://cdn.jsdelivr.net/gh/starfederation/datastar@main/bundles/datastar.js` everywhere, otherwise things can easily get stale. The `@main` branch will only ever be updated with official versions.

rmcsharry:
> Great work @eduwass! No further comments from me.
> 
> For what it's worth, as a non-maintainer, looks good to me. 👍

Same sentiments here! 

@jmstevers I think you can merge this now.

sdk/typescript/examples/node/node.js
eduwass: OK, updated @bencroker ! 👍

ghost:
LGTM @bencroker

ghost (approved):

bencroker (approved):

bencroker:
Merged. HUGE thanks for this, great work!!

eduwass:
Awesome!! 🚀

Thanks everyone 👌

PR #986, opened 2025-07-11
Python SDK: Prevent sanic error log with datastar_response

gazpachoking:
The `datastar_response` decorator for sanic was returning the `DatastarResponse` even if it had already streamed the whole response from a generator. Everything worked, but sanic would log an error when the response was returned. This fixes that.


bencroker (approved):

bencroker:
Thanks!

PR #991, opened 2025-07-11
Feature: clojure-SDK - build scripts to publish to clojars

JeremS:
With RC1 here, it is time to release the SDK to clojars instead of asking our user to use it as a git dependency with the need to track the right commit.

This PR adds to the work of @Ramblurr by providing build scripts to 2 other libraries that work with the SDK. There are no changes to the SDK.

In `sdk/clojure/doc/maintainers-guide.md` there are release instructions for the core team under the `Release` section.
Once the clojars account is setup, there is github action that must be triggered manually to build and publish a new version.

Cheers,


sdk/clojure/sdk-malli-schemas/build.clj
```diff
@@ -0,0 +1,49 @@
...
+  (:require [clojure.tools.build.api :as b]
+            [clojure.edn :as edn]))
+
+(def project (-> (edn/read-string (slurp "deps.edn"))
+                 :aliases :neil :project))
+(def lib (or (:name project) 'my/lib1))
+
+;; use neil project set version 1.2.0 to update the version in deps.edn
+
+(def version (or (:version project)
```
Ramblurr: I like to remove this `or` defaults on version and lib, they make no sense and if something goes wrong you don't want to release an artifact with those dummy values.

(Same for the other build.clj)

Ramblurr (changes requested):

sdk/clojure/sdk-malli-schemas/build.clj
JeremS: You are right the neil generated default isn't that great. I can remove the the dummy artifact names. If no name is found I could exit 1 or just throw. Which is cleaner though? If there is an error here will the github workflow just fail and stop?

JeremS: Scratch my last comment, you already had setup the proper behavior in the other scripts. I replicated it.

Ramblurr (approved):

JeremS:
@bencroker Hey Ben, I think we can close this PR, I have integrated the code for this in the new datastar-clojure repo.

bencroker:
Great, thanks!

PR #998, opened 2025-07-13
initial cleanup and JSON definitions

delaneyj:


delaneyj:
We are doing this more piecemeal in #1007

PR #1002, opened 2025-07-14
Ruby support for Phlex 2x

ismasan:
Fix support for https://www.phlex.fun components.

Phlex is a popular library for server-rendered HTML components in Ruby.

Phlex 2 changed its API from `#call(view_context)` to `#call(context:)`.

```ruby
class Title < Phlex::HTML
  def initialize(title:)
    @title = title
  end

  def view_template
    h1(id: 'title') { @title }
  end
end

datastar.patch_elements Title.new(title: 'Hello!')
```


bencroker (approved):

bencroker:
Thanks!

PR #1006, opened 2025-07-15
Fix go sdk tests

gazpachoking:
Switch the -server flag over to the tests runner, where it was used
Make sure the embedfs worked on Windows by always using a forward slash for paths.

I don't Go, so this is liable to be wrong, but seems to work for me now.


delaneyj:
This is included in #1007

PR #1007, opened 2025-07-15
Go SDK tests

ismasan:
Take just `sdk/tests` from `repo-per-sdk` branch, plus @gazpachoking 's fixes.

The idea is to make these Go tests available on their own, so that SDKs can rely on them as they move into their own repos.

To test locally, you can

1). Launch your test app, for example on `http://localhost:8000`
2). In this repo: `cd sdk/tests`
3). `go run ./cmd/datastar-sdk-tests -server http://localhost:8000`

For reference, this is how this can be used during CI on each SDK's repo https://github.com/starfederation/datastar-ruby/pull/1


delaneyj (commented):
I think we can remove the Docker stuff completely now

delaneyj (approved):
Talking in the Discords this is a better approach than my scorched earth version

bencroker:
Merged, thanks all!!

PR #1025, opened 2025-07-24
Tweak how keys are parsed from the data-* attribute

gazpachoking:
Originally just investigating why css variables weren't able to be set by data-style (`data-style---myvar="blah"`) from this [thread](https://discord.com/channels/1296224603642925098/1397653251364622366) I found a few areas that could be cleaned up which allowed that and other possibilities. Since we are not trying to translate any arbitrary piece of text to any arbitrary case we don't need to do as much. We know the input is dataset case, (which for simple things is the same as camel case) and we know that it was translated from the attribute string, which is just kebab case for simple things. Rather than trying to sanitize the input any further we can just use those two cases.

## Overview of new functionality:
- dataset casing is totally reversible, make case.kebab mean the raw (valid) string directly from the attribute (a superset of traditional kebab casing)
- dataset casing is a superset of camel case. Pass the key as parsed by the dataset through for camel case.
- pascal casing is just dataset case with a first capital letter

## Breaking stuff
Some of these changes are breaking, but mostly only if you were putting extraneous characters in key names and relying on d* to sanitize them for you (at least I think those are the only cases.) Here are some of the ones I've seen:
- `data-signals-a1a` used to become a1A, now becomes a1a. (You can achieve the old value with `data-signals-a1-a`)
- `data-signals-a1a__kebab` used to become a-1-a. Now becomes a1a (`data-signals-a-1-a__kebab` stays the same though, which seems saner than asking the casing mod to add the dashes)

## New stuff
Things that you couldn't do before this change (when using key form):
- Start a key with a number. `data-signals-1foo='bar'`
- Have a number in the midst of all lower case letters `data-signals-lower2lower`
- Start a key with a capital (without pascal case). `data-signals--foo='bar'` (I would never do this, but don't see why it should be barred.)
- Use the literal form from the raw attribute when it had e.g. multiple dashes in a row. `data-signals---foo__case.kebab='bar'` (this is the case that also allows css vars in data-style `data-style---foo='bar'`
- Mix in numbers without capitalizing things. `data-signals-a2a='bar'` would become `a2A` before (not sure what the purpose of this was)
- Use an alias that had a dash in. (It would strip the wrong length alias, since dataset turns it to camel case)

I ran the `* casing` tests from the test pages, along with some manual tests, and everything seemed to come out okay, but I could have missed some things.

I'll put some comments on the code inline below explaining.

> your PR is totally whack.
> – delaneyj

### DIFF
--- a/library/src/engine/engine.ts
+++ b/library/src/engine/engine.ts
@@ -1,6 +1,6 @@
 import { isHTMLOrSVG } from '../utils/dom'
 import { isEmpty, isPojo, pathToObj } from '../utils/paths'
-import { camel, snake } from '../utils/text'
+import { camel, lowerFirst, snake } from '../utils/text'
 import { DATASTAR, DSP, DSS } from './consts'
 import { initErr, runtimeErr } from './errors'
 import type {
@@ -836,7 +836,7 @@ export function load(...pluginsToLoad: DatastarPlugin[]) {
     return a.name.localeCompare(b.name)
   })
 
-  pluginRegexs = plugins.map((plugin) => RegExp(`^${plugin.name}([A-Z]|_|$)`))
+  pluginRegexs = plugins.map((plugin) => RegExp(`^${plugin.name}([A-Z-]|_|$)`))
 }
 
 function applyEls(els: Iterable<HTMLOrSVG>): void {
@@ -875,7 +875,7 @@ function applyAttributePlugin(
   attrKey: string,
   value: string,
 ): void {
-  const rawKey = camel(alias ? attrKey.slice(alias.length) : attrKey)
+  const rawKey = alias ? lowerFirst(attrKey.slice(camel(alias).length)) : attrKey
   const plugin = plugins.find((_, i) => pluginRegexs[i].test(rawKey))
   if (plugin) {
     // Extract the key and modifiers
@@ -883,7 +883,7 @@ function applyAttributePlugin(
 
     const hasKey = !!key
     if (hasKey) {
-      key = camel(key)
+      key = lowerFirst(key)
     }
     const hasValue = !!value
 
--- a/library/src/engine/engine.ts
+++ b/library/src/engine/engine.ts
@@ -1,6 +1,6 @@
 import { isHTMLOrSVG } from '../utils/dom'
 import { isEmpty, isPojo, pathToObj } from '../utils/paths'
-import { camel, lowerFirst, snake } from '../utils/text'
+import { camel, removePrefix, snake } from '../utils/text'
 import { DATASTAR, DSP, DSS } from './consts'
 import { initErr, runtimeErr } from './errors'
 import type {
@@ -875,16 +875,13 @@ function applyAttributePlugin(
   attrKey: string,
   value: string,
 ): void {
-  const rawKey = alias ? lowerFirst(attrKey.slice(camel(alias).length)) : attrKey
+  const rawKey = alias ? removePrefix(alias, attrKey) : attrKey
   const plugin = plugins.find((_, i) => pluginRegexs[i].test(rawKey))
   if (plugin) {
     // Extract the key and modifiers
-    let [key, ...rawModifiers] = rawKey.slice(plugin.name.length).split(/__+/)
+    let [key, ...rawModifiers] = removePrefix(plugin.name, rawKey).split(/__+/)
 
     const hasKey = !!key
-    if (hasKey) {
-      key = lowerFirst(key)
-    }
     const hasValue = !!value
 
     // Create the runtime context
--- a/library/src/utils/text.ts
+++ b/library/src/utils/text.ts
@@ -4,18 +4,19 @@ export const isBoolString = (str: string) => str.trim() === 'true'
 
 export const kebab = (str: string) =>
   str
-    .replace(/([a-z0-9])([A-Z])/g, '$1-$2')
-    .replace(/([a-z])([0-9]+)/gi, '$1-$2')
-    .replace(/([0-9]+)([a-z])/gi, '$1-$2')
+    .replace(/([A-Z])/g, '-$1')
     .toLowerCase()
 
 export const camel = (str: string) =>
-  kebab(str).replace(/-./g, (x) => x[1].toUpperCase())
+  kebab(str).replace(/-[a-z]/g, (x) => x[1].toUpperCase())
 
 export const snake = (str: string) => kebab(str).replace(/-/g, '_')
 
 export const pascal = (str: string) =>
   camel(str).replace(/(^.|(?<=\.).)/g, (x) => x[0].toUpperCase())
+
+export const lowerFirst = (str: string) =>
+    (str[0] === "-" ? "" : str[0].toLowerCase()) + str.slice(1)
 
 export const jsStrToObject = (raw: string) => {
   try {
--- a/library/src/utils/text.ts
+++ b/library/src/utils/text.ts
@@ -7,16 +7,20 @@ export const kebab = (str: string) =>
     .replace(/([A-Z])/g, '-$1')
     .toLowerCase()
 
-export const camel = (str: string) =>
-  kebab(str).replace(/-[a-z]/g, (x) => x[1].toUpperCase())
+export const camel = (str: string) => str
 
 export const snake = (str: string) => kebab(str).replace(/-/g, '_')
 
 export const pascal = (str: string) =>
-  str[0].toUpperCase() + camel(str).slice(1)
+  str[0].toUpperCase() + str.slice(1)
 
-export const lowerFirst = (str: string) =>
-    (str[0] === "-" ? "" : str[0].toLowerCase()) + str.slice(1)
+export const removePrefix = (prefix: string, str: string) => {
+  // removes a prefix from a string in dataset case
+  // e.g. removePrefix("signals", "signalsKeyname") -> "keyname"
+  const suffix = str.slice(prefix.replace(/-[a-z]/g, "-").length)
+  if (!suffix) return suffix
+  return (suffix[0] === "-" ? "" : suffix[0].toLowerCase()) + suffix.slice(1)
+}
 
 export const jsStrToObject = (raw: string) => {
   try {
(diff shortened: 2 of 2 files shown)


library/src/engine/engine.ts
```diff
@@ -875,15 +875,15 @@ function applyAttributePlugin(
   attrKey: string,
   value: string,
 ): void {
-  const rawKey = camel(alias ? attrKey.slice(alias.length) : attrKey)
```
gazpachoking, line 878: This was calling camel just to lowercase the first letter. Calling camel messed things up when there were literal dashes in the key though.

library/src/engine/engine.ts
```diff
@@ -875,15 +875,15 @@ function applyAttributePlugin(
...
+  const rawKey = alias ? lowerFirst(attrKey.slice(camel(alias).length)) : attrKey
   const plugin = plugins.find((_, i) => pluginRegexs[i].test(rawKey))
   if (plugin) {
     // Extract the key and modifiers
     let [key, ...rawModifiers] = rawKey.slice(plugin.name.length).split(/__+/)
 
     const hasKey = !!key
     if (hasKey) {
-      key = camel(key)
```
gazpachoking, line 886: Same as above. Since the key follows a dash that isn't meant to capitalize the next letter, but separate the key from the plugin name, we just need to lower case the first letter.

library/src/utils/text.ts
```diff
@@ -4,19 +4,20 @@ export const isBoolString = (str: string) => str.trim() === 'true'
 
 export const kebab = (str: string) =>
   str
-    .replace(/([a-z0-9])([A-Z])/g, '$1-$2')
-    .replace(/([a-z])([0-9]+)/gi, '$1-$2')
-    .replace(/([0-9]+)([a-z])/gi, '$1-$2')
+    .replace(/([A-Z])/g, '-$1')
```
gazpachoking, line 7: I'm not quite sure what all of this was for. This is where I potentially broke something, but it seemed fine in my tests. I just made this the inverse of the camel case function so things round trip.

library/src/utils/text.ts
```diff
@@ -4,19 +4,20 @@ export const isBoolString = (str: string) => str.trim() === 'true'
...
-  kebab(str).replace(/-./g, (x) => x[1].toUpperCase())
+  kebab(str).replace(/-[a-z]/g, (x) => x[1].toUpperCase())
 
 export const snake = (str: string) => kebab(str).replace(/-/g, '_')
 
 export const pascal = (str: string) =>
   camel(str).replace(/(^.|(?<=\.).)/g, (x) => x[0].toUpperCase())
 
+export const lowerFirst = (str: string) =>
+    (str[0] === "-" ? "" : str[0].toLowerCase()) + str.slice(1)
```
gazpachoking: When we use a dash in the middle of a data-* attribute we need to undo the camel casing on only the first character. That's usually lower casing the first character, unless the first character was a dash, in which case that means the real first character of our key was a non-alpha character. In that case we just strip the dash off.

library/src/engine/engine.ts
```diff
@@ -836,7 +836,7 @@ export function load(...pluginsToLoad: DatastarPlugin[]) {
     return a.name.localeCompare(b.name)
   })
 
-  pluginRegexs = plugins.map((plugin) => RegExp(`^${plugin.name}([A-Z]|_|$)`))
+  pluginRegexs = plugins.map((plugin) => RegExp(`^${plugin.name}([A-Z-]|_|$)`))
```
gazpachoking, line 839: We want plugins to match `<pluginname>-` from the raw string, which normally becomes the plugin name followed by an upper case character after the dataset camelcasing, except in the instance that the first character of the key was not a letter, then it will stay a -.

library/src/engine/engine.ts
```diff
@@ -875,15 +875,15 @@ function applyAttributePlugin(
   attrKey: string,
   value: string,
 ): void {
-  const rawKey = camel(alias ? attrKey.slice(alias.length) : attrKey)
+  const rawKey = alias ? lowerFirst(attrKey.slice(camel(alias).length)) : attrKey
```
gazpachoking: This was calling camel just to lowercase the first letter. Calling camel messed things up when there were literal dashes in the key though. Also, it was only needed in the case that we have an alias. Called camel here on the alias as well in case there was a dash in there the length will have changed.

library/src/engine/engine.ts
```diff
@@ -875,15 +875,15 @@ function applyAttributePlugin(
...
   const plugin = plugins.find((_, i) => pluginRegexs[i].test(rawKey))
   if (plugin) {
     // Extract the key and modifiers
     let [key, ...rawModifiers] = rawKey.slice(plugin.name.length).split(/__+/)
 
     const hasKey = !!key
     if (hasKey) {
-      key = camel(key)
+      key = lowerFirst(key)
```
gazpachoking: Same as above. Since the key follows a dash that isn't meant to capitalize the next letter, but separate the key from the plugin name, we just need to lower case the first letter.

library/src/engine/engine.ts
gazpachoking: ~~On second thought the alias should probably just be camelized once when it's set rather than every time we are matching here.~~ Oh right, we still need it the normal way still when aliasifying.

library/src/utils/text.ts
gazpachoking: lowerFirst is a bad name. What this does is removes the dash after the plugin name (e.g. data-signals**-**), but since dataset has already done it's camel thing the dash may have turned into a capital letter. This leaves us with what would have been in the dataset had the key been directly after the `data-` rather than after our plugin name. Useful both for getting the raw key name, as well as getting the plugin name when it occurs after an alias.

library/src/utils/text.ts
```diff
@@ -4,18 +4,23 @@ export const isBoolString = (str: string) => str.trim() === 'true'
...
   str
-    .replace(/([a-z0-9])([A-Z])/g, '$1-$2')
-    .replace(/([a-z])([0-9]+)/gi, '$1-$2')
-    .replace(/([0-9]+)([a-z])/gi, '$1-$2')
+    .replace(/([A-Z])/g, '-$1')
     .toLowerCase()
 
-export const camel = (str: string) =>
-  kebab(str).replace(/-./g, (x) => x[1].toUpperCase())
+export const camel = (str: string) => str
```
gazpachoking, line 10: The dataset already did enough work of camelizing. Just pass it through. There are some differences, but it's just that the original camel function stripped more things that the user never would have put in the attribute string if they were sane and wanted camel case.

library/src/utils/text.ts
```diff
@@ -4,18 +4,23 @@ export const isBoolString = (str: string) => str.trim() === 'true'
...
-  camel(str).replace(/(^.|(?<=\.).)/g, (x) => x[0].toUpperCase())
+  str[0].toUpperCase() + str.slice(1)
+
+export const removePrefix = (prefix: string, str: string) => {
+  // removes a prefix from a string in dataset case
+  // e.g. removePrefix("signals", "signalsKeyname") -> "keyname"
+  const suffix = str.slice(prefix.replace(/-[a-z]/g, "-").length)
+  if (!suffix) return suffix
+  return (suffix[0] === "-" ? "" : suffix[0].toLowerCase()) + suffix.slice(1)
+}
```
gazpachoking, line 23: Combines the work of stripping a prefix and changing the first letter of the remaining string to be if it hadn't had a hyphen in front. Also handles the case for stripping an alias that has a dash. (but doesn't have a dash anymore in the dataset case string)

gazpachoking:
Here's the page I've been using to test. Signal values are what the signal name comes out to with this PR.
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
<!--    <script type="module" src="/dist/bundles/datastar.js"></script>-->
    <script type="module" src="https://cdn.jsdelivr.net/gh/starfederation/datastar@main/bundles/datastar.js"></script>
</head>
<body>
    <div data-works-the-same
         data-signals="{aoeu: 'aoeu'}"
         data-signals-foo="'foo'"
         data-signals-foo-bar="'fooBar'"
         data-signals-foo-bar__case.pascal="'FooBar'"
         data-signals-foo-bar__case.snake="'foo_bar'"
         data-signals-fooo1="'fooo1'"
         data-does-not-currently-work
         data-signals-1-foo-bar="'1FooBar'"
         data-signals--f-o-o="'FOO'"
         data-signals-1foo="'1foo'"
         data-signals---foo="'-Foo'"
         data-signals---foo__case.kebab="'--foo'"
         data-signals-1foo-bar__case.pascal="'1fooBar'"
         data-behavior-changed
         data-signals-a1a="'a1a'"
         data-signals-c1c__case.kebab="'c1c'"
         data-signals-foo1bar__case.kebab="'foo1bar'"
         data-signals-b-1-b="'b-1B'"
         data-signals-d-1-d__case.kebab="'d-1-d'"
         data-signals-foo-1="'foo-1'"
         data-signals-dash--in--camel__case.camel="'dash-In-Camel'"
         data-ref="testDiv"
    ></div>
    <h3>The raw html:</h3>
    <pre data-text="$testDiv.outerHTML.replaceAll(' ', '\n')"></pre>
    <h3>The resultant signals:</h3>
    <pre data-json-signals></pre>
    <h3>el.dataset:</h3>
    <pre data-text="JSON.stringify($testDiv.dataset, null, 2)"></pre>
</body>
</html>

```

library/src/utils/text.ts
bencroker: Doesn’t this break the `__case.camel` modifier when used in `data-class-dark-blue__case.camel`?

library/src/engine/engine.ts
bencroker: Seems valid.

library/src/utils/text.ts
bencroker: I suppose your changes assume the key is coming in as dataset casing.

gazpachoking: Yep, all the casing mods now assume dataset casing input. (And I think I eliminated all use of them outside that)

gazpachoking:
I missed multiple things here. There are multiple uses of the casing functions that come from different source cases other than the case mods. There are multiple places that expect you can convert casing multiple times (i.e. `kebab(camel(x))`) I still think there are some subtle bugs and inconsistencies that can be fixed by knowing the source and destination case and doing less rather than using general case changing functions, but this PR is not it yet.

PR #1029, opened 2025-07-29
Add PocketBase SDK

benallfree:


bencroker:
The SDK folders are going away, and are only here for legacy purposes. I can add the SDK to the site at https://data-star.dev/reference/sdks. Can it go under TypeScript?

benallfree:
I think it’s more akin to Craft CMS or Laravel in the sense that PocketBase is a runtime. PocketPages and the SDK will not run in a node environment, for example. 

So probably just a PocletBase heading makes the most sense to me.

PR #1032, opened 2025-07-31
Delay file input events until bound signal has updated

gazpachoking:
If you bind a signal to a file input, and also have a change/input listener on that input, the event handler will get called before the signal has been updated with the new data. [From this thread](https://discord.com/channels/1296224603642925098/1399474846261575721)

The following will log 'undefined' the first time the change event is fired, and be one change behind for subsequent events.
```html
<input type="file" data-bind-file data-on-change="console.log($fileNames[0])">
```

This fixes that by stopping the propagation of the event in the bind handler and re-dispatching it once the signal has been updated. Feels hacky though.

### DIFF
--- a/library/src/plugins/attributes/bind.ts
+++ b/library/src/plugins/attributes/bind.ts
@@ -78,7 +78,10 @@ export const Bind: AttributePlugin = {
           }
           break
         case 'file': {
-          const syncSignal = () => {
+          const syncSignal = (evt: Event) => {
+            const target = evt.target
+            if ('detail' in evt || target === null) return
+            evt.stopImmediatePropagation()
             const files = [...(el.files || [])]
             const contents: string[] = []
             const mimes: string[] = []
@@ -119,6 +122,7 @@ export const Bind: AttributePlugin = {
                   },
                 ),
               )
+              target.dispatchEvent(new CustomEvent(evt.type, {detail: true}))
             })
           }
 


library/src/plugins/attributes/bind.ts
```diff
@@ -119,6 +122,7 @@ export const Bind: AttributePlugin = {
                   },
                 ),
               )
+              target.dispatchEvent(new CustomEvent(evt.type, {detail: true}))
```
gazpachoking, line 125: Hmm, maybe this should go in a `finally` so the event doesn't get swallowed in case there is an error reading the files.

bencroker:
This feels like something that should be handled in userland rather than in the `bind` plugin. Do you agree?

gazpachoking:
I'm not sure. How are you thinking it should be solved in user land?

bencroker:
It depends on what you’re trying to do?

gazpachoking:
Say you are trying to post a file signal to the backend on a change event. Or even just log it like in the PR description.

bencroker:
You could delay the the expression to ensure it runs after the signal is synced, for example.

```html
<input type="file" data-bind-file data-on:change__delay.100ms="...">
```

bencroker:
Presumably the same issue applies to other input types too, which is why I think this should be left to the user to handle.

gazpachoking:
File is the only one which is set async, so it's the only one with this issue. Using an arbitrary delay isn't so nice, because there is no guarantee how long it might take (though I suspect even large files wouldn't take _that_ long, so that may not be an issue in practice.) `data-effect` or `data-on-signal-patch` can work to work around the issue, which is good enough, it's just confusing the first time you run into the problem because it works different from everything else. I don't feel strongly that it must be fixed, just unfortunate UX that it doesn't work the same as everything else.

bencroker:
Yeah I see what you mean. But doesn’t your change rely on attribute order, since event listeners are executed in the order they were added? In other words, the following would _not_ work.

```html
<input type="file" data-on:change="..." data-bind-file>
```

gazpachoking:
I believe you are correct about the order, but that would be similarly broken for all input types

bencroker:
For some reason this PR feels hacky to me. Technically the `change` event has fired, and I don’t think we should suppress it. 

I’m more inclined to add a dedicated event that users can listen for. That way, there’s no question about what the event means or when it is dispatched.

```js
.finally(() => {
  el.dispatchEvent(new CustomEvent('datastar-bind-file', { bubbles: true }))
})
```

What do you think?

gazpachoking:
I agree this PR feels hacky, I just wasn't sure a better way to do it. I guess if we are conceding that bound files are going to work different than everything else (with regard to hooking the change event) the new event is cleaner than doing a data-on-signal-patch with a data-on-signal-patch-filter to filter it down to just the bound signal (which would be functionally the same if I understand correctly.)

I'm not totally sure what's the best approach is, because I'm just imagining scenarios and not solving an actual problem with an app at the moment. The disadvantage of the file binding not being congruent is you can't handle many inputs with a bubbled event.
i.e. You must change this:
```html
<div data-on:change="@post('/somewhere')">
  <input type="text" data-bind-text />
  <input type="text" data-bind-text2 />
  <input type="file" data-bind-file />
</div>
```
to this:
```html
<div>
  <input type="text" data-bind-text data-on:change="@post('/somewhere')"/>
  <input type="text" data-bind-text2 data-on:change="@post('/somewhere')"/>
  <input type="file" data-bind-file data-on:datastar-bind-file="@post('/somewhere')"/>
</div>
```

gazpachoking:
Maybe just documenting that it works differently is best. :shrug:

bencroker:
I think you’re conflating the `change` event, which is dispatched by the browser, with a signal being modified, which is performed by Datastar. If it is signal patches you want to listen to, then `data-on-signal-patch` is the appropriate attribute to use.

```html
<div data-on-signal-patch="@post('/somewhere')">
```

gazpachoking:
> I think you’re conflating the `change` event, which is dispatched by the browser, with a signal being modified, which is performed by Datastar. If it is signal patches you want to listen to, then `data-on-signal-patch` is the appropriate attribute to use.

I get the distinction. Since we can't use the change event to act on file inputs bound to a signal (since the signal won't be updated yet) the workaround would be to act when the signal updates. I just meant that your proposed `datastar-bind-file` event ends up being the same as data-on-signal-patch (filtered down to just the one signal), because it's explicitly being fired at the same time the file signal is updated.

PR #1041, opened 2025-08-18
data-bind creates single signal for file inputs

gazpachoking:
# Proposal
data-bind should create a single signal for 'file' type inputs with a list of objects describing the files.

This would better match how the data-bind plugin works for all other input types. Every other input type only creates one signal, the exact one the user has specified to bind to. Currently file inputs result in two extra signals that the user has not explicitly specified.

This would also better match the underlying `FileList` api that the input element provides (Which is a list of `File` objects.) I have named the mime field `type` to match the [File interface](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/file#type)

PR with implementation is here, but no need to use it, just wanted to put this proposal up formally somewhere.

### DIFF
--- a/library/src/plugins/attributes/bind.ts
+++ b/library/src/plugins/attributes/bind.ts
@@ -79,9 +79,7 @@ export const Bind: AttributePlugin = {
         case 'file': {
           const syncSignal = () => {
             const files = [...(el.files || [])]
-            const contents: string[] = []
-            const mimes: string[] = []
-            const names: string[] = []
+            const signalFiles: object[] = []
             Promise.all(
               files.map(
                 (f) =>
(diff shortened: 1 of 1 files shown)


library/src/plugins/attributes/bind.ts
```diff
@@ -79,9 +79,7 @@ export const Bind: AttributePlugin = {
         case 'file': {
           const syncSignal = () => {
             const files = [...(el.files || [])]
-            const contents: string[] = []
-            const mimes: string[] = []
-            const names: string[] = []
+            const signalFiles: object[] = []
```
ghost:
```suggestion
type File = {
  name: string
  type: string
  contents: string
}
const signalFiles: File[] = []
```

gazpachoking:
This also opens up the possibility to include other fields from `File` e.g. size

bencroker:
Thanks for the PR! We’ll look into getting this added for the next release.

bencroker:
We added an implementation based on yours for the next release, thanks again!

screenshot-9AA69h9x@2x

bencroker:
Released in [v1.0.0-RC.6](https://github.com/starfederation/datastar/releases/tag/v1.0.0-RC.6), thanks @gazpachoking!

PR #1045, opened 2025-08-19
Cancel requests when element is removed

wmTJc9IK0Q:
This makes it so that when requestCancellation is `auto` and the element that initiated the request is removed, the request will now be cancelled. Otherwise open SSE requests keep getting new events and patching elements that shouldn't be and resources are not released causing memory leaks.

### DIFF
--- a/library/src/engine/types.ts
+++ b/library/src/engine/types.ts
@@ -54,7 +54,7 @@ export type WatcherPlugin = {
 }
 
 export type ActionPlugins = Record<string, ActionPlugin>
-export type ActionMethod = (ctx: RuntimeContext, ...args: any[]) => any
+export type ActionMethod = (ctx: ActionContext, ...args: any[]) => any
 
 export type ActionPlugin = {
   type: 'action'
@@ -99,6 +99,10 @@ export type RuntimeContext = InitContext & {
   fnContent?: string // the content of the function
   evt?: Event // the event that triggered the plugin
   runtimeErr: (reason: string, metadata?: object) => Error
+}
+
+export type ActionContext = RuntimeContext & {
+  setCleanup: (fn: OnRemovalFn, key: string) => void // sets a cleanup function for this element and key
 }
 
 export type RuntimeExpressionFunction = (
--- a/library/src/engine/engine.ts
+++ b/library/src/engine/engine.ts
@@ -4,6 +4,7 @@ import { camel, snake } from '../utils/text'
 import { DATASTAR, DSP, DSS } from './consts'
 import { initErr, runtimeErr } from './errors'
 import type {
+  ActionContext,
   ActionPlugins,
   AttributePlugin,
   Computed,
@@ -1000,17 +1001,21 @@ function applyAttributePlugin(
 
       const cleanup = plugin.onLoad(ctx)
       if (cleanup) {
-        let cleanups = removals.get(el)
-        if (cleanups) {
-          cleanups.get(rawKey)?.()
-        } else {
-          cleanups = new Map()
-          removals.set(el, cleanups)
-        }
-        cleanups.set(rawKey, cleanup)
+        setCleanup(cleanup, el, rawKey);
       }
     }
   }
+}
+
+function setCleanup(cleanup: OnRemovalFn, el: HTMLOrSVG, key: string): void {
+  let cleanups = removals.get(el)
+  if (cleanups) {
+    cleanups.get(key)?.()
+  } else {
+    cleanups = new Map()
+    removals.set(el, cleanups)
+  }
+  cleanups.set(key, cleanup)
 }
 
 // Set up a mutation observer to run plugin removal and apply functions
@@ -1160,6 +1165,7 @@ function generateReactiveExpression(
   const actionMatches = [...expr.matchAll(actionsRe)]
   const actionNames = new Set<string>()
   const actionFns = new Set<(...args: any[]) => any>()
+  const actionCtx: ActionContext = { ...ctx, setCleanup: (cleanup: OnRemovalFn, key: string) => setCleanup(cleanup, ctx.el, key) }
   if (actionMatches.length) {
     const actionPrefix = `${DATASTAR}Act_`
     for (const match of actionMatches) {
@@ -1175,7 +1181,7 @@ function generateReactiveExpression(
       // Add ctx to action calls
       expr = expr.replace(`@${actionName}(`, `${name}(`)
       actionNames.add(name)
-      actionFns.add((...args: any[]) => action.fn(ctx, ...args))
+      actionFns.add((...args: any[]) => action.fn(actionCtx, ...args))
     }
   }
 
--- a/library/src/plugins/backend/actions/fetch.ts
+++ b/library/src/plugins/backend/actions/fetch.ts
@@ -48,6 +48,7 @@ export const createHttpMethod = (
 
     if (!isDisabled && !(requestCancellation instanceof AbortController)) {
       fetchAbortControllers.set(el, controller)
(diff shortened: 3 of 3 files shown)


library/src/engine/types.ts
```diff
@@ -101,6 +101,10 @@ export type RuntimeContext = InitContext & {
   runtimeErr: (reason: string, metadata?: object) => Error
 }
 
+export type ActionContext = RuntimeContext & {
+  setCleanup: (fn: OnRemovalFn, key: string) => void // sets a cleanup function for this element and key
```
wmTJc9IK0Q, line 105: Since actions can return values to callers, I needed to add a new context API that allow setting a cleanup based on a key. The key can't be auto-set to the action name because there could be different fetch actions used on the same element and we only want to keep the removalFn for the last fetch action.

romshark:
I faced a problem today ([see Discord thread](https://discord.com/channels/1296224603642925098/1407452673866731652)) and it seems this change is addressing it.

I wanted to *bind* an SSE connection to a particular view, and once that view is morphed away or replaced the SSE connection would reset. Currently it just keeps creating more SSE connections on every page transition. I kept wondering why my app is bugging out, turns out it was the multiple SSE connections patch-racing.

wmTJc9IK0Q:
@wmTJc9IK0Q Yes that is similar to my use case. I have an SSE endpoint that powers a specific server-side component and pushes updates to it. Whenever the parent component removes this child component I want the requests to stop and no more patching of elements.

romshark:
Maybe, for reasons I currently don't see, it makes sense to allow the old behavior through an option still:
```
<div id="t" data-on-load="@get('/stream', {detach: true})"></div>
```
Detaching means it won't get canceled if `t` is removed from the DOM.

But it definitely makes sense to make cancelation the new default behavior. If you want to keep an SSE stream triggered by an element, keep that element. Current behavior is rather counterintuitive and potentially dangerous (memory leaks, patch races).

bencroker:
Thanks for the PR. We need to discuss internally if and how we want to handle this.

wmTJc9IK0Q:
@bencroker  Anything I can do to move this forward?

bencroker:
Not really. We've been busy and haven't come to a decision on this yet.

bencroker:
@wmTJc9IK0Q since you’re requesting a breaking change in behaviour, can you please explain and justify the argument for it? Specifically, what problem is it supposed to solve, and what is preventing you from solving it using an `AbortController` with the [`requestCancellation`](https://data-star.dev/reference/actions#requestCancellation) option?

goertzenator:
I could really use this.  Are there any work-arounds?

goertzenator:
If I may, consider a UI with a navigation bar and main display area such that clicking on a navbar element `@get`s new content for the main display area.  When the main display area is replaced, you want all connections associated with the old area to be destroyed; this is how HTMX behaves including the SSE and Websocket extensions.  It feels intuitive, modular, and coherent.  You want this for the same reason that you want an existing connection cancelled when a new one is launched, which is currently the default.  If you don't do this, you get long running connections piling up and making a mess.  

Approaching this from the other direction: Is the current behavior desirable?  Are people really designing apps that depend on connections persisting once the element that launched it is gone?

So, I like this PR and I think most people will want and expect this behavior.

wmTJc9IK0Q:
@bencroker I see two scenarios where this can be useful, depending on how the app is being architected:

## Single connection per page

In the often "recommended" pattern of having a single connection for each page for all element and signal updates, if I want to navigate to a new page without a full page refresh, the old page connection needs to be closed or else invalid updates will keep getting pushed to the browser window and leave old requests open, leaking server and browser resources.

This is the pattern I am currently using. As it stands right now, I have to do a full-page refresh which is slower and requires doing a lot of extra work to persist UI state. For example, I have global UI elements for filter state and a time range picker. I want that to persist while navigating through the app. So I have 3 options:
1. Encode all of UI state into query params in ALL the hrefs on the page (or put everything in local storage, but now I have more state to deal with) and do full page navs. This is a worse UX and I also lose other important state, such as scroll position, open state on `<details>` tags, etc. I would need to add a lot more signals and boilerplate to persist/load all of this to local storage or URL params.
2. Make it so that all links to other pages handle request cancellation of the open request - littering the page with boilerplate for the "recommended" architecture. I can add a function for this but I still have to remember to do it everywhere and if I forget there will be issues.
3. Have datastar automatically close open requests for removed elements (what this PR does). My app doesn't have to do anything.

## Component-specific connections

As my team grows on the project, I want to enable developers to write their own isolated "components". Some of those may have their own ways of loading data and one way to solve that could be to have them open their own SSE connections for fetching new data without touching the main page's SSE endpoint. 

In this case, if those components get removed from the page by a navigation or some change in a higher component in the main page's tree I would want the requests to be closed on those leaf elements.

This case is more theoretical at this point for me. I was building things this way at first but was able to create a pattern on my backend for doing one request for the whole page that is working fine for async component updates right now.

---

I have been using my datastar build from this PR in production for 2 months now and I have not encountered any issues with this behavior as of yet.

I do have the same question as @goertzenator - I don't see how the existing behavior is at all desirable. I would add to that: what will break because of this? If anything this feels like a bug in the auto cancellation feature and I'd be surprised that anyone is relying on this behavior on purpose. If they are - I'd like to understand why.

The auto cancel feature was also just recently added and I think now is the time to fix this bug before it goes stable.

wmTJc9IK0Q:
Edited option (1) about how local storage could be used instead of URL params, but this has it's own problems as well.

bencroker:
Thanks for the PR and discussion! We’ve taken a different approach that has the same result, coming in RC.6.

wmTJc9IK0Q:
@bencroker  

Ok, I’ll be glad if my desired behavior is there, but I don’t like being asked for a justification, spending time writing a very detailed response and then not getting any engagement back on that from you. So I wouldn’t call it a discussion really. 

I get being annoyed with thankless open source users but I’m an actual engaged dev committing some of my time and experience to your project. You are spending plenty of time answering random discord threads, so I don’t get the lack of engagement to someone who is actually reading and trying to improve your code. 

Guess that’s just not as fun as telling people you’re using it wrong?

delaneyj:
> @bencroker
> 
> Ok, I’ll be glad if my desired behavior is there, but I don’t like being asked for a justification, spending time writing a very detailed response and then not getting any engagement back on that from you. So I wouldn’t call it a discussion really.

If you don't like justifying changes, this isn't the project for you.  

> I get being annoyed with thankless open source users but I’m an actual engaged dev committing some of my time and experience to your project. You are spending plenty of time answering random discord threads, so I don’t get the lack of engagement to someone who is actually reading and trying to improve your code.

We read your thoughts, thanked you and discussed internally the ramifications to you.... and everyone else.  Your time caused engagement with the ideas and even a robust response that WE will support, not you.

> Guess that’s just not as fun as telling people you’re using it wrong?

No, they decided that making a solution that we feel comfortable was simpler and a better use of time than trying to coax this PR along.  Your PR is a suggestion, not a mandate.  Neat part of open source is you get zero say on how someone spends there time or PRs or Discord.  If you don't like that fork and support your own approach.

bencroker:
Released in [v1.0.0-RC.6](https://github.com/starfederation/datastar/releases/tag/v1.0.0-RC.6).

wmTJc9IK0Q:
Yeah you totally missed my point. There was no mandate, I was asking for engagement from you on my lengthy explanation and ideas. You aren't going to be successful running an open source project if you're treating people like this.

You all don't know how to engage in fruitful discussions, so I'm no longer going to engage or contribute to this. Thanks.

delaneyj:
Oh well good luck on your future endeavors, hopefully you find a project that is run exactly like you want.

Issue #1056, opened 2025-09-07
Option to skip form validation


thomersch:
### Feature Request

Background: I got a form that contains a list of lines. A button on the form allows to add another line, something like `<input type="button" data-on-click="@get('/line/', {contentType: 'form'})">`, this sends the current form, and the backend responds with the lines with an empty line at the end.

Problem: The form isn't valid (yet) and Datastar always calls `checkValidity`, so the browser just displays an error instead of sending the get request.

Possible solution: [HTML allows a novalidate attribute on forms](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form#novalidate). Datastar could check if this is set and omit validation for such forms.

(I am not sure that my approach of adding lines is the most idiomatic way, so if anyone got suggestions, give them to me.)

bencroker:
If you just want to add another line, `data-on-click="@get('/line/')"` should work without checking validity.

thomersch:
This makes sense, but unfortunately it's a bit cumbersome in my case, because I use Django forms which need line numbers. I could track the current line number using a signal, but there is also a management form that needs the current line count that I'd need to modify. And when I delete a line, I'd need to renumber them to be consecutive. Thus I decided it would be easier to just redraw all the lines when adding or removing.

Would you be open to a PR that would contain a check for novalidate?

bencroker:
Well, what input fields are failing validation, and couldn’t you remove their validation requirements instead?

thomersch:
Input fields that are not part of the lines and are only relevant to the outer document when it's submitted for saving. I'd say it'd be much easier to toggle the validation requirement on the form instead of having to disable it on all form elements.

HTMX does respect this setting: https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L3590

bencroker:
Yeah I suppose we can add a check for the [novalidate](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form#novalidate) attribute. I’ll make a PR in our dev repo.

bencroker:
Added for the RC.6 release.

bencroker:
Released in [v1.0.0-RC.6](https://github.com/starfederation/datastar/releases/tag/v1.0.0-RC.6).

stianhj:
Hey, doesn't seem like the novalidate fix was actually included in RC6? I have form that has novalidate that still gets checked before submission.

bencroker:
[...]

PR #1083, opened 2025-10-30
Respect novalidate in form submission

stianhj:

## Problem Statement

This issue doesn't seem to be actually resolved? https://github.com/starfederation/datastar/issues/1056

## Proposed Solution

Added a simple check for if the form element has the novalidate attribute.


bencroker:
Thanks but we already have a solution for this, it just didn’t make it in as explained in https://github.com/starfederation/datastar/issues/1056#issuecomment-3467696017.

PR #1091, opened 2025-11-06
Emit the last event from a throttle interval when using 'trailing'

gazpachoking:
Based on [this bug report from discord](https://discord.com/channels/1296224603642925098/1435388700489023620).

## Problem Statement
- When using the throttle modifier with the 'trailing' option, the expression gets executed with the event from the leading edge of the throttle window rather than the last event in that throttle window.

e.g. if you had data-on:keydown__throttle.1s.noleading.trailing and you typed 'abcde' all within a second, you'd expect your expression to get the 'e' event at the end of the throttle window, but instead you get the 'a' event. 

- If using throttle or debounce with both leading and trailing edges, a single event would be multiplied into two events. Since the point of debounce/throttle is to _reduce_ the quantity of events, this doesn't make sense.

## Proposed Solution

Store the latest event while waiting in a throttle window to emit when 'trailing' option is enabled.

Don't emit a trailing event when both leading and trailing are enabled if there was only one event within the window.

Both of these behaviors now match how lodash handle debounce/throttle.

I also unified the two timing functions into one since it seemed easier than maintaining the two separate implementations.

### DIFF
--- a/library/src/utils/timing.ts
+++ b/library/src/utils/timing.ts
@@ -12,60 +12,44 @@ export const delay = (
   }
 }
 
+export const throttle = (
+  callback: EventCallbackHandler,
+  wait: number,
+  leading = true,
+  trailing = false,
+  debounce = false,
+): EventCallbackHandler => {
+  let lastArgs: Parameters<EventCallbackHandler> | null = null
+  let timer = 0
+
+  return (...args: any[]) => {
+    if (timer) {
+      lastArgs = args
+    }
+    if (leading && !timer) {
+      callback(...args)
+      lastArgs = null
+    }
+    if (!timer || debounce) {
+      timer && clearTimeout(timer)
+      timer = setTimeout(() => {
+        if (trailing && lastArgs !== null) {
+          callback(...lastArgs)
+        }
+        lastArgs = null
+        timer = 0
+      }, wait)
+    }
+  }
+}
+
 export const debounce = (
   callback: EventCallbackHandler,
   wait: number,
   leading = false,
   trailing = true,
 ): EventCallbackHandler => {
-  let timer = 0
-  return (...args: any[]) => {
-    timer && clearTimeout(timer)
-
-    if (leading && !timer) {
-      callback(...args)
-    }
-
-    timer = setTimeout(() => {
-      if (trailing) {
-        callback(...args)
-      }
-      timer && clearTimeout(timer)
-      timer = 0
-    }, wait)
-  }
-}
-
-export const throttle = (
-  callback: EventCallbackHandler,
-  wait: number,
-  leading = true,
-  trailing = false,
-): EventCallbackHandler => {
-  let waiting = false
-  let lastArgs: Parameters<EventCallbackHandler> | null = null
-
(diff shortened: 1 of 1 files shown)


gazpachoking:
I expanded the problem to include a single event within a window being multiplied into two events when both leading and trailing edges were enabled. Rather than fix that problem separately for the debounce function I just combined them into one function with a debounce option.

bencroker:
Thanks for the PR, it looks really good! Interestingly, you chose to put most of the logic in the `throttle` function, whereas Lodash puts in the `debounce` function:
https://github.com/lodash/lodash/blob/933e1061b8c344d3fc742cdc400175d5ffc99bce/lodash.js#L10965-L10981

Personally, I’d prefer a dedicated internal helper function (something like `rateLimit`) that both functions call. This would make it explicit that they share rate-limiting logic, rather than one being built on top of the other (even if makes the bundle a few bytes larger).

gazpachoking:
> Interestingly, you chose to put most of the logic in the `throttle` function, whereas Lodash puts in the `debounce` function

I tried it reversed, but did it this way for naming reasons. 'debounce' felt like a special case of throttling whereas the reverse didn't seem true. (And having a function called debounce that took a parameter to turn debouncing off seemed crazy. :laughing:)

> Personally, I’d prefer a dedicated internal helper function (something like `rateLimit`) that both functions call. This would make it explicit that they share rate-limiting logic, rather than one being built on top of the other (even if makes the bundle a few bytes larger).

Happy to separate that out if you prefer, 'throttle' and 'rate limit' sorta feel like synonyms to me though.

gazpachoking:
Or wait, do we not even bother making a 'throttle' and 'debounce' function? We can just make a rateLimit function and call it with the appropriate arguments from the modifyTiming function.

bencroker:
> Or wait, do we not even bother making a 'throttle' and 'debounce' function? We can just make a rateLimit function and call it with the appropriate arguments from the modifyTiming function.

Yes, calling it directly from the `modifyTiming` function feels like the right approach. On second thought, however, “throttling” is technically more correct than “rate limiting”, so we can  just remove the `debounce` function altogether.

bencroker:
Just tested using https://data-star.dev/tests/on_click_debounce and debounce appears to not be working. How are you testing your implementation?

gazpachoking:
I was just testing with an on key down for a text input. I did make that last change blind though, I'll test again in a bit and against the example you have

gazpachoking:
> Just tested using https://data-star.dev/tests/on_click_debounce and debounce appears to not be working. How are you testing your implementation?

Yep, I had a bug when there was only a single input in the timing window and only trailing edge was enabled. Thought I was testing all the cases, but turns out I missed that one.
Separately, seems like the window could be a little longer on that debounce test. At 100ms I am not capable of clicking fast enough to actually get the debounce behavior to kick in.

bencroker (approved):

bencroker:
Merged, thank you!

bencroker:
FYI I added some tests and changed the timing of them to 500ms.

bencroker:

PR #1094, opened 2025-11-11
Remove `remove` mode being able to omit `selector` from the ADR

ghost:

## Problem Statement

SDK ADR contains outdated information regarding the `remove` mode being able to omit `selector` and specify the element to remove via a top level id.

## Proposed Solution

Update the ADR.

## Alternatives Considered

None.


bencroker (approved):

bencroker:
Merged, thanks!

Issue #1104, opened 2025-12-11
Potential miscompilation with dollar sign in string literal inside expression


yurivish:
### Bug Report

Repro: https://observablehq.com/d/9c297f594cb1895a with  Datastar 1.0.0-RC.6 in Safari 26.1 on macOS, and Chrome 143.0.7499.41.

In short,
```html
<div data-text='"$testing"'></div>
```
renders as `$['testing']` and
```html
<div data-text="'$testing'"></div>
```
throws an error.

I didn't see this specific limitation in the docs (https://data-star.dev/guide/datastar_expressions/)  I found out about it after I added console.log statements of the form
```js
console.log('$width: ', $width)
```
which were transformed into invalid JavaScript and thus did not work.

### Datastar Version

1.0.0-RC.6

bencroker:
Datastar converts `$testing` to `$['testing']`, since the dollar sign represents a signal in Datastar expressions, which explains why you’re seeing the output you are.

yurivish:
I think it is fine for Datastar to convert `$testing` into `$['testing']`, but surprising for Datastar to convert `"$hi"` into `"$['hi']"`, or `'$hi'` into `'$['hi']'`, which is a syntax error. 

You're more knowledgeable than I am here, but to me it seems like this might be a limitation of the expression parser.

If this is intended behavior to keep the parser simple, I think that's a choice one can make, but the documentation should document those cases where valid JavaScript will not be accepted (the second case), or will have a different than intended effect (the first case).

bencroker:
> to me it seems like this might be a limitation of the expression parser.

There is no expression parser, just regex replace: 
https://github.com/starfederation/datastar/blob/421724a8c185b859c58c94a2a585068a27473d25/library/src/engine/engine.ts#L370-L383

yurivish:
In my opinion, if this behavior is intended then it should be documented. Would you agree?

Also, is there any input that would trigger the third regex replace (on line 380)? I think the case in the comment on line 379 ($ inside brackets) is taken care of by the previous replacement on line 374, and I haven't found an input string that would cause the third replace call to replace anything.

Edit: [Fuzz test link](https://observablehq.com/d/06769bf07de4162e)

bencroker:
> In my opinion, if this behavior is intended then it should be documented. Would you agree?

Maybe. We can’t document _everything_. Feel free to make a suggestion.

> Also, is there any input that would trigger the third regex replace (on line 380)?

[...]

PR #1105, opened 2025-12-12
Remove unnecessary regex replacement

yurivish:
## Problem Statement

I noticed that the expression-processing code has what appears to be a regex replacement that is not necessary, since the case it was intended to handle is already handled by the previous replacements.

## Proposed Solution

Remove the `replace` call.

## Alternatives Considered

Do not remove the call, leaving the redundant code as-is, with small but nonzero impacts on runtime performance and bundle size.

---

## Analysis 

This is the code in question, where the third regex has no apparent effect:

```js
  expr = expr
    // $['x'] -> $x (normalize existing bracket notation)
    .replace(/\$\['([a-zA-Z_$\d][\w$]*)'\]/g, '$$$1')
    // $x -> $['x'] (including dots and hyphens)
    .replace(/\$([a-zA-Z_\d]\w*(?:[.-]\w+)*)/g, (_, signalName) =>
      signalName
        .split('.')
        .reduce((acc: string, part: string) => `${acc}['${part}']`, '$'),
    )
    // [$x] -> [$['x']] ($ inside brackets)
    .replace(
      /\[(\$[a-zA-Z_\d]\w*)\]/g,
      (_, varName) => `[$['${varName.slice(1)}']]`,
    )
```

The third regex would only be needed if there were a $identifier inside brackets that the second regex somehow missed. But the second regex handles that case, and an [exhaustive fuzz test](https://observablehq.com/d/06769bf07de4162e)  of all possible 8-character strings containing the characters `x$.-0[]ø ` shows that there are no strings of that length on which the third replacement has any effect.

If this analysis is correct, then this PR will have no behavioral changes.

### DIFF
--- a/library/src/engine/engine.ts
+++ b/library/src/engine/engine.ts
@@ -376,11 +376,6 @@ const genRx = (
         .split('.')
         .reduce((acc: string, part: string) => `${acc}['${part}']`, '$'),
     )
-    // [$x] -> [$['x']] ($ inside brackets)
-    .replace(
-      /\[(\$[a-zA-Z_\d]\w*)\]/g,
-      (_, varName) => `[$['${varName.slice(1)}']]`,
-    )
 
   expr = expr.replaceAll(/@(\w+)\(/g, '__action("$1",evt,')
 


bencroker:
Thanks for the PR but this was already acknowledged in https://github.com/starfederation/datastar/issues/1104#issuecomment-3647010626 and has since been removed for the upcoming RC.7.

yurivish:
Great. Your comment there was that this needed further testing, which is why I made this PR and included the results of the expanded fuzz test.

bencroker:
Again, thank you for the effort! I tested the change against our test suite and all tests passed.

yurivish:
Yep, thanks. Will ask before making a PR next time.

bencroker:
https://github.com/starfederation/datastar/blob/develop/CONTRIBUTING.md#pull-requests

yurivish:
I saw that, but had assumed that our discussion in the issue meant it made sense to open a PR. Clearly my intuition is off when it comes to this project.

bencroker:
And I assumed the issue was a proposal to improve the docs 😄

Issue #1088, opened 2025-11-05
Inserting elements into SVG doesn't work, wrong NamespaceURI


ecraven:
### Bug Report

If you insert any element (via @patch for example) anywhere inside an svg tag, it shows up in the DOM inspector, but it isn't rendered, because the namespaceURI of the element is incorrect ("http://www.w3.org/1999/xhtml"
instead of "[http://www.w3.org/2000/svg"](http://www.w3.org/2000/svg%22)). 

My guess is that datastar always uses `document.createElement`, but it would have to use `document.createElementNS` for anything that goes inside SVG.

Is there anything I can do locally to fix this, or would I have to wait for a patch in datastar proper?

Here is a reproduction of it, after 2 seconds, the rectangle is patched and disappears, because the namespace is incorrect. https://codepen.io/cpuser543/pen/jEWRgxv. After the rectangle disappears, go to the inspector, select the rectangle, and look at its `namespaceURI`

Thank you for a great library!

### Datastar Version

v1.0.0-RC.6

bencroker:
I’m not sure we can do anything about this, but I’ll leave it to @jmstevers to respond in full.

@ecraven please see the [SVG morphing example](https://data-star.dev/examples/svg_morphing) if you haven’t already.

ghost:
I don't think this is generally solveable as html and svg elements share names (e.g. `<a>` and `<a>`) and the new elements that are being patched in are passed into each patch mode and the namespace of an element is immutable. It may be possible to just check if parentElement is svg namespace then recreate all the new elements with svg namespace but im not sure if that would work 100% of the time. This would have to work for every patch mode.

ecraven:
well, right now, this means you just can't use datastar to work with SVG (unless you never modify the structure), which is not an ideal situation either :-/ the correct solution is probably what you mentioned, check the parent element's namespace (that might suffice, maybe there's no need to go up and search for an explicit `svg` element). I'm not sure whether html can be nested inside svg again, etc...
it would be nice if this could be fixed in some way :-/

yurivish:
Would it be crazy to think about adding a `namespace` option to [PatchElements](https://github.com/starfederation/datastar/blob/develop/sdk/ADR.md)?

bencroker:
> well, right now, this means you just can't use datastar to work with SVG

@ecraven have you seen the [SVG Morphing example](https://data-star.dev/examples/svg_morphing)?

ecraven:
[...]

PR #1108, opened 2025-12-14
Add namespace support to patchElements

yurivish:

## Problem Statement

There is currently no way to control the namespace of patched elements, which prevents insertion of new elements into `<svg>` and `<math>` contexts where children must be created in the appropriate XML namespace.

This PR came out of a discussion in issue https://github.com/starfederation/datastar/issues/1088.

## Proposed Solution

Add a new option, `wrap`, to `datastar-patch-elements`, which specifies a tag name (such as `svg` or `math`) that is used as a wrapper element when parsing the elements to be patched. The wrapper element is never inserted into the document, and is used only to influence the namespace of the elements being inserted by the patch operation.

I'm not sure about the name `wrap`, since the wrapping technique is an implementation detail and the wrapper element does not appear in the document. I considered calling it `namespace` but saying `data: namespace div` feels _weird_. Maybe `namespaceTag`?

## Alternatives Considered

One alternative considered was to wrap patched elements individually in `<svg>` or `<math>` tags as a way to set their namespace, but this comes at a performance and memory [cost](https://github.com/starfederation/datastar/issues/1088#issuecomment-3649631675) since the wrapper elements persist in the document after patching. The cost of these wrapper elements can be significant when using SVG for data visualization since it's not uncommon to have thousands of elements (eg. circles in a scatterplot) that you might want to individually add/animate/remove.

---

<details>
  <summary>Full proposal</summary>
<br />

The [HTML spec](https://html.spec.whatwg.org/) specifies `<svg>` and `<math>` as the only allowed "foreign elements", and the namespace of children is determined by these parent tags.

The proposal is to add an optional patch-elements argument, `wrap`, which can be used to specify the name of a wrapper tag during element creation, allowing the namespace URI to be set while avoiding the costs associated with inserting wrapper elements into the document itself. 

When using `wrap` Datastar will construct one extra wrapper element per patch call (which can of course patch multiple elements at once), which is in addition to the `<body>` and `<template>` elements that Datastar already typically creates. The wrapper element is never placed into the document and is only used to influence the namespaces of its children.

Examples:

```
event: datastar-patch-elements
data: mode append
data: selector #vis
data: wrap svg
data: elements <circle id="c1" cx="10" r="5" fill="red"/>
data: elements <circle id="c2" cx="20" r="5" fill="green"/>
data: elements <circle id="c3" cx="30" r="5" fill="blue"/>
```

```
event: datastar-patch-elements
data: selector #equation
data: mode inner
data: wrap math
data: elements <mi>x</mi><mo>=</mo><mfrac><mn>1</mn><mn>2</mn></mfrac>
```

Since elements can now be created in multiple namespaces, there is also a small change/bugfix to morphing to inherit the namespace from the existing element when creating a new empty child to morph into, ie. keeping the namespaces of any elements that already exist.

</details>

### DIFF
--- a/library/src/plugins/actions/fetch.ts
+++ b/library/src/plugins/actions/fetch.ts
@@ -237,6 +237,7 @@ type ResponseOverrides =
       selector?: string
       mode?: string
       useViewTransition?: boolean
+      namespace?: string
     }
   | {
       onlyIfMissing?: boolean
--- a/library/src/plugins/actions/fetch.ts
+++ b/library/src/plugins/actions/fetch.ts
@@ -560,6 +560,7 @@ const fetchEventSource = (
             'selector',
             'mode',
             'useViewTransition',
+            'namespace'
           )
         }
 
--- a/library/src/plugins/watchers/patchElements.ts
+++ b/library/src/plugins/watchers/patchElements.ts
@@ -23,14 +23,14 @@ type PatchElementsArgs = {
   mode: PatchElementsMode
   selector: string
   useViewTransition: boolean
-  wrap?: string
+  wrap: string
 }
 
 watcher({
   name: 'datastar-patch-elements',
   apply(
     ctx,
-    { elements = '', selector = '', mode = 'outer', useViewTransition, wrap },
+    { elements = '', selector = '', mode = 'outer', useViewTransition = '', wrap = '' },
   ) {
     switch (mode) {
       case 'remove':
@@ -54,7 +54,7 @@ watcher({
       mode,
       selector,
       elements,
-      useViewTransition: useViewTransition?.trim() === 'true',
+      useViewTransition: useViewTransition.trim() === 'true',
       wrap,
     }
 
--- a/library/src/plugins/watchers/patchElements.ts
+++ b/library/src/plugins/watchers/patchElements.ts
@@ -23,13 +23,14 @@ type PatchElementsArgs = {
   mode: PatchElementsMode
   selector: string
   useViewTransition: boolean
+  wrap?: string
 }
 
 watcher({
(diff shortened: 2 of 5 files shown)


bencroker:
This looks great! I made some minor tweaks for the sake of consistency.

yurivish:
@bencroker oops – sorry, I just force-pushed over your changes with a docs update 🤦‍♂️. Was using force push to keep the commit history clean. Can you please re-apply your changes?

Edit: Looks like I have them locally; applying…

bencroker:
Having discussed it internally and reviewing the ADR changes, I think `namespace` would be more suitable than `wrap`, unless you see value in being able to wrap the elements in an arbitrary tag? The optional `namespace` option would accept a string enum `'html' | 'svg' | 'mathml'`, defaulting to `html` in `patchElements`.

yurivish:
I agree that's cleaner. How's this?

bencroker:
Much better! I refactored to use const arrays, and added a generic `isValidType` function that acts as a type guard, so that both `PatchElementsMode` and `Namespace` types can be checked in the same way. Any last tweaks you want to make?

yurivish:
Thanks, looks good to me. I just made two more tweaks, one prompted by a [comment](https://github.com/starfederation/datastar/issues/1088#issuecomment-3655883059) from @ecraven to allow the namespace option to be used with `text/html` responses, and another to update the sdk spec json files with the new option.

bencroker:
Added the `namespace` option to `ResponseOverrides` and merged. Excellent work, thank you!

yurivish:
🎉

bencroker:

Issue #1111, opened 2025-12-18
VSCode extension breaks highlighting for Laravel Blade files


icylace:
### Bug Report

The Datastar VSCode extension v1.3.0 breaks highlighting for my Laravel Blade files.

In this screenshot, line 5, the one with the special attribute highlighting, breaks highlighting that follows it:
<img width="1032" height="591" alt="Image" src="https://github.com/user-attachments/assets/b33cfc67-9c3b-41e1-ba0b-528f457739a2" />

When I comment out that line, highlighting for the rest of the file goes back to normal:
<img width="980" height="593" alt="Image" src="https://github.com/user-attachments/assets/c6842ca6-fc77-4505-bb97-b2c0d6e29d8f" />

Here is how it looks with the extension turned off:
<img width="984" height="594" alt="Image" src="https://github.com/user-attachments/assets/28e1c5d5-3585-4ec6-88ab-65e4187ffd1d" />

### Datastar Version

Datastar VSCode extension v1.3.0

bencroker:
@YuryKL could you please take a look?

YuryKL:
Looks like the injection selector was a bit permissive and included `L:text.html` which is any HTML text context, not just inside tags. Removing it and keeping L:meta.tag it should restrict it to just inside html tags and prevent it bleeding into php blocks. If I could get a gist of a file that breaks it currently I can do a more thorough test but that should work as a fix.

icylace:
@YuryKL Sure, here is a gist of the file in its entirety: https://gist.github.com/icylace/9ff018477b0da0a514e88e7c2f8ab70f

YuryKL:
Great. Tested without `L:text.html` and it seems like its no longer bleeding into php stuff.

<img width="862" height="700" alt="Image" src="https://github.com/user-attachments/assets/f001572a-cb8c-482a-beee-d553af21a4ab" />

<img width="942" height="237" alt="Image" src="https://github.com/user-attachments/assets/4bb47db1-f37c-4e6c-aeec-258651ec375e" />

bencroker:
Fixed by https://github.com/starfederation/datastar/pull/1112 and release in 1.3.1, thanks!

PR #1112, opened 2025-12-19
Fix injectionSelector in grammar to prevent leaking outside of HTML Tags

YuryKL:

## Problem Statement
Addressing the issue in https://github.com/starfederation/datastar/issues/1111

## Proposed Solution
Remove the injectionSelector `L:text.html` while leaving `L:meta.tag` so that the grammar does not activate outside of html tags.


bencroker:
Merged, thanks!

PR #1114, opened 2025-12-23
Fix minor typescript error

oceantume:
This line of code fails the typescript compiler check.

## Problem Statement

Running the `tsc` CLI on the project fails with `Duplicate identifier namespace. (ts 2300)`.

## Proposed Solution

Remove the duplicate identifier


bencroker:
Merged, thanks!

PR #1123, opened 2026-01-17
Fix data received by delayed on-signal-patch

### LINKED ISSUE #1122: data-on-signal-patch when debounced receives the wrong list of changes
## The example

I was trying to create a page with auto-save and an indicator.

**Signals**
```js
{
    data: { /* stuff to save */ },
    version: 0, // updated by the frontend on every change
    savedVersion: 0, // updated by the backend confirming the version saved.
}
```

**auto-save**
```html
<div data-on-signal-patch__debounce.500ms="@post(..)"
      data-on-signal-patch-filter="{include: /^data/}"
></div>
```

**indicator**
```html
<div data-on-signal-patch="$version++"
    data-on-signal-patch-filter="{include: /^data/}"
    data-text="($version === $savedVersion) ? 'saved' : 'saving'"
></div>
```

## What happens

1. User clicks on something, `{ data: .. }` is updated
2. The non-debounced on-signal-patch runs immediately and receives changes `{ data: .. }`. It sets { version: 1 }
3. The debounced on-signal-patch runs after its delay, but receives changes { version: 1 } and the filter is applied, so the value isn't evaluated.

## What I expected

I expected the debounced on-signal-patch to react to { data: .. } but that is not what it received.
I also notice that `evt` is undefined when the value is run, which is not entirely unexpected.

### Datastar Version

v1.0.0-RC.7

jpeg729:

## Problem Statement

The on-patch-signals plugin with modified timing will sometimes not fire when combined with filters. When a second patch signals event occurs before the value is executed, the event is replaced, which is good unless the new event doesn't match the filters.

## Proposed Solution

Filter events before applying throttling.

## Alternatives Considered

We could leave it as is and propose workarounds based on data-on-interval or similar, but the fix seems straightforward and the upside seems clear.

Fixes #1122

### DIFF
--- a/library/src/plugins/attributes/onSignalPatch.ts
+++ b/library/src/plugins/attributes/onSignalPatch.ts
@@ -44,8 +44,7 @@ attribute({
     const callback: EventListener = (evt: Event | CustomEvent<JSONPatch>) => {
       // we know that evt is CustomEvent<JSONPatch>, but typescript doesn't
       const watched = filtered(filters, (evt as CustomEvent<JSONPatch>).detail)
-      if (!isEmpty(watched))
-      {
+      if (!isEmpty(watched)) {
         timedCallback({...evt, detail: watched})
       }
     }
--- a/library/src/plugins/attributes/onSignalPatch.ts
+++ b/library/src/plugins/attributes/onSignalPatch.ts
@@ -29,25 +29,26 @@ attribute({
       filters = jsStrToObject(filtersRaw)
     }
 
-    let running = false
-
-    const callback: EventListener = modifyTiming(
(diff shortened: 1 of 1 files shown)


jpeg729:
I removed the `running` flag because as far as I can see it doesn't do anything useful. It is set to true, then to false, inside a block of code that runs synchronously in a single-threaded environment.

Maybe you prefer a proper type check on the type of the event, but I think I am right in saying that the old code didn't so a type assertion is not worse.

jpeg729:
Here is a more minimalist example https://codepen.io/jpeg729/pen/OPXmZML

PR #1128, opened 2026-01-30
Morph improvements

MichaelWest22:
Here is a possible improvement to datastars morph algorithim.

Posting this PR here so it can be assessed and tested before possibly merging.

I've included a commit that creates compiled bundles changes to make initial testing easier but these need to be removed before we merge something like this

Also included a minor fix to disable id="" from persistent id's that is an edge case that can break morph.

There are a couple of scan ahead limit variables set to 5 and 10 nodes which could be adjusted.


bencroker:
Thanks for the PR! We’ll review this internally and do some testing.

bencroker:
Sorry about the delay in responding. We had our best man on this, @andersmurphy, but he seems to be permanently on holiday these days. He did encounter some issues while testing, so hopefully he can enlighten us as to what those were, so we can move forward with this PR.

PR #1138, opened 2026-03-10
Morph attributes and update values

### LINKED ISSUE #1075: inconsistent interpretation of html spec in handling of input "checked" attribute
### Bug Report

- datastar special cases the "value" attribute of input elements; when patching elements, if an `<input>` has a changed "value", a corresponding change is made to the input's value _property_.
- datastar does not similarly special case the "checked" attribute of input elements; when patching elements, if an input's "checked" attribute has changed, the input's "checked" _property_ is not updated.
- "value" and "checked" are similarly defined in the html spec (https://html.spec.whatwg.org/#attr-input-value); the spec suggests no reason d* should handle them differently.

A change that that would resolve this inconsistency:
- datastar should special-case the input attributes "checked" and "disabled" the same way it currently special cases "value".
- datastar should also special case the `<option>` "selected" attribute, which is described in the html spec similarly to input.value and input.checked.
- in case it's helpful, this file shows what elements and attributes morphdom handles specially: https://github.com/patrick-steele-idem/morphdom/blob/master/src/specialElHandlers.js

These changes would simplify the HTML included in some patch-elements events. Instead of toggling a checkbox with:
```html
<input type="checkbox" data-on-load="el.checked = true;" />
```

users could write:
```html
<input type="checkbox" checked />
```

This also seems like a more intuitive behavior. It seems unlikely someone would send down `<input type="checkbox" checked />` and only want to change the default checkedness of the input.

### Datastar Version

v1 rc 5

bencroker:
This PR morphs attributes and updates values based on incoming DOM element attributes. If attributes do not match the _original_ values, they are modified.

It can be tested by importing the bundle at: 
https://github.com/starfederation/datastar/blob/morph-attributes-update-values/bundles/datastar.js

Closes https://github.com/starfederation/datastar/issues/1075


gazpachoking:
A few remaining issues:

For checkboxes:
The checked attribute does indeed update, but the checked state of the box doesn't.

For select options:
The users selection is overwritten with the selected item on morph no matter what. (looks like it should check `newOptions[i].getAttribute('selected') !== oldOptions[i].getAttribute('selected')` rather than the `selected` property.)

For everything:
I think it should probably emit the 'change' event on the element when value/checked/selected gets updated, so that data-bind signals stay in sync when the element is updated by morph.

bencroker:
Good catches! Addressed in https://github.com/starfederation/datastar/pull/1138/changes/cd10a280608b2593b6fc7c342fd9c35580d3b3a6.

library/src/plugins/watchers/patchElements.ts
```diff
@@ -565,31 +573,46 @@ const morphNode = (
...
         oldElt.value = newElt.getAttribute('value') ?? ''
+        shouldDispatchChangeEvent = true
       }
     } else if (
       oldElt instanceof HTMLTextAreaElement &&
       newElt instanceof HTMLTextAreaElement
     ) {
-      if (newElt.value !== oldElt.value) {
+      // Modify only if the new element’s value is different from the old element’s initial value.
+      if (oldElt.firstChild && oldElt.firstChild.nodeValue !== newElt.value) {
```
gazpachoking: This won't update if the original element didn't have any text, but text is morphed in. Maybe better to just switch to oldElt.textContent, so that we don't have to dance around oldElt maybe not having a child?

library/src/plugins/watchers/patchElements.ts
```diff
@@ -598,11 +621,23 @@ const morphNode = (
     ).split(' ')
 
     for (const { name, value } of newElt.attributes) {
```
gazpachoking, line 624: Hmm, Unchecking a checkbox via morph still isn't working. That case gets skipped in this loop because the 'checked' attribute is not present on the newElt. It does get caught in the next loop and the attribute is removed, but during that iteration no properties are updated.

gazpachoking (commented):
Seems very close now! Couple more edge cases I noticed.

library/src/plugins/watchers/patchElements.ts
bencroker: Fixed in https://github.com/starfederation/datastar/pull/1138/changes/b6a86d74f78c3881be4e188618e3add55dc3c63c. My tests weren’t catching it because we also morph nodes later in the code.

library/src/plugins/watchers/patchElements.ts
bencroker: Should be fixed now.

bencroker:
Refactored and simplified the logic to allow regular attribute morphing to handle non-special cases.

gazpachoking:
Seems good to me now! Nice work! :rocket:

PR #1156, opened 2026-04-23
Prevent cloning element when unused

### LINKED ISSUE #1155: Extra deep-clone when materializing a string?
### Bug Report

I just happened to notice this during debugging, but it seems like an unnecessary performance hit, if I'm understanding the code correctly.

In patchElements.ts, in function `onPatchElements`, if `elements` is a string, `consume` is set to false.  Then, when `applyToTargets` is called, if `consume` is false, the new DOM is deep-cloned.  It seems to me that, if the elements came from a string, they should not be cloned.

### Datastar Version

v1.0.0

bencroker:
Fixes https://github.com/starfederation/datastar/issues/1155.


delaneyj (approved):
> This is a DOM API constraint but, nevertheless, 🖕JS.

Pure cinema

PR #1160, opened 2026-04-30
Bind selector bugfix

### LINKED ISSUE #1159: checkbox/radio input issues with data-bind with modifiers
### Bug Report

In the case that multiple checkboxes are bound to an array signal, having a modifier on the data-bind breaks the binding completely. This is due to data-bind using a [selector](https://github.com/starfederation/datastar/blob/d3beb4bb1f6c24613386408a8c629a9176000c41/library/src/plugins/attributes/bind.ts#L99) to find all the checkboxes bound to the same signal which doesn't take into account that the data-bind attribute might have modifiers.
Codepen: https://codepen.io/gazpachoking/pen/NPbWBde

There is a similar though less critical issue with radio buttons. When a group of radio buttons are bound to an uninitialized signal, they initialize the signal to the value of the checked radio button (if there is one.) The [selector](https://github.com/starfederation/datastar/blob/d3beb4bb1f6c24613386408a8c629a9176000c41/library/src/plugins/attributes/bind.ts#L75) to find the checked radio button fails similarly if the data-bind attributes have modifiers. The signal remains uninitialized, and none of the radio buttons start checked.
Codepen: https://codepen.io/gazpachoking/pen/Jobjmdz

### Datastar Version

v1.0.1

bencroker:
Fixes https://github.com/starfederation/datastar/issues/1159.


gazpachoking:
I see, you went with the method of requiring all the elements bound to the same input to have the exact same data-bind attribute format. Not the perfect solution, but seems a reasonable expectation.

bencroker:
Yes. There was an additional bug in the original version in which a key of `foo-bar` would end up as a value selector of `="foo-bar"` instead of `="fooBar"`. So I decided to simplify with the assumption that all binds for the same signal should use the same syntax.

PR #1162, opened 2026-05-04
fix(fetch): Apply backoff to HTTP error retries

### LINKED ISSUE #1161: `retryMaxCount` and `retryScaler` are ignored when `retry='error'` and response is `500`
### Bug Report

**Reproduction:** https://jsbin.com/vomomofezu/1/edit?html,console,output
**Expected Behavior**: I expect settings such as `retryMaxCount`, `retryScaler` to apply for all retry cases. If this behavior is expected then the [action docs](https://data-star.dev/reference/actions) must define it and explain the rationale.

Tip: `502 bad gateway` is fine, but `500` and `503` are not.

romshark:
fixes https://github.com/starfederation/datastar/issues/1161

## Problem Statement

As described in https://github.com/starfederation/datastar/issues/1161

## Proposed Solution

Use the same retry scheduling for both error paths.
Make the HTTP-error branch also dispatch `retrying` same as the network error path.

## Alternatives Considered

Either eimply duplicate, without `scheduleRetry`?

```
diff --git a/library/src/plugins/actions/fetch.ts b/library/src/plugins/actions/fetch.ts
index 91a789e9..39bf37d6 100644
--- a/library/src/plugins/actions/fetch.ts
+++ b/library/src/plugins/actions/fetch.ts
@@ -571,6 +571,12 @@ const fetchEventSource = (
           ) {
             clearTimeout(retryTimer)
             retryTimer = setTimeout(create, retryInterval)
+            retryInterval = Math.min(retryInterval * retryScaler, retryMaxWait)
+            if (++retries >= retryMaxCount) {
+              dispatchFetch(RETRIES_FAILED, el, {})
+              dispose()
+              reject('Max retries reached.')
+            }
             return
           }
           dispose()
```

Or if that behavior is actually expected - the reasoning should be stated in the [action docs](https://data-star.dev/reference/actions)

### DIFF
--- a/library/src/plugins/actions/fetch.ts
+++ b/library/src/plugins/actions/fetch.ts
@@ -520,6 +520,18 @@ const fetchEventSource = (
 
     let retries = 0
     let baseRetryInterval = retryInterval
+    const scheduleRetry = (interval: number = retryInterval): boolean => {
+      clearTimeout(retryTimer)
+      retryTimer = setTimeout(create, interval)
+      retryInterval = Math.min(retryInterval * retryScaler, retryMaxWait)
+      if (++retries >= retryMaxCount) {
+        dispatchFetch(RETRIES_FAILED, el, {})
+        dispose()
+        reject('Max retries reached.')
+        return false
+      }
+      return true
+    }
     const create = async () => {
       curRequestController = new AbortController()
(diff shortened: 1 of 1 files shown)


romshark:
Closing since you've now got https://github.com/starfederation/datastar/pull/1163

PR #1179, opened 2026-07-02
Optimize regex using non-capturing group

bencroker:
Optimizes the regex used to split statements in the engine to prevent unnecessary backtracking.

Thanks to @nickchomey for providing this suggestion via Discord.


library/src/engine/engine.ts
nickchomey: Interestingly, the `?:`  are used here already. Though, I see one or two groups where it could perhaps be added?

nickchomey:
p.s. i just ran the benchmark script that i shared with your new version of the regex (with the couple extra `?:`), and it is 58% faster whereas what I proposed was 45% faster.  Seems to pass all the tests i had in the script. 

Will still probably only result in 0.1ms of savings on page renders though! Still worth doing

library/src/engine/engine.ts
bencroker: Good catch! I added another non-capturing group, and improved the replacement regex while at it.

bencroker:
Yeah, a micro-optimisation, but an optimisation nevertheless. I took the opportunity to clean up some code, too!

PR #1184, opened 2026-07-08
fix: data-attr not removing old keys

Regaez:
## Problem Statement

If the `data-attr` value object previously included keys that _no longer exist_, those old attributes remain present on the element, instead of being removed. This can cause issues when dynamically generating attribute keys via signals.

[View this codepen for live demos of the following examples](https://codepen.io/Regaez/pen/dPNdVbZ)

### Example 1

The `data-on-interval` attribute is dynamically applied with its `__duration` based off a signal. Every time `data-on-interval` runs, it increments `$i` by 1, meaning it _should_ wait an additional second each time before running again, and `$i` should only ever increase by 1.
    
```html
<div data-signals="{ i: 1 }">
  <span data-text="$i"></span>
  <progress max="100" data-bind:i data-attr="{ [`data-on-interval__duration.${$i}s`]: '$i++' }"></progress>
</div>
```

### Example 2

Only the currently selected event type _should_ be attached to the input via `data-on`, upon which it will trigger custom validation of the input.
```html
<label>
  Validate on
  <select data-signals:event="'change'" data-bind:event>
    <option value="change">change</option>
    <option value="input">input</option>
    <option value="focus">focus</option>
    <option value="blur">blur</option>
  </select>
</label>
<input 
  type="text" 
  data-attr="{ [`data-on:${$event}`]: `el.setCustomValidity('Validated on: ' + $event); el.reportValidity();` }" 
/>
```

## Proposed Solution

Cache the previous object keys that were applied by `data-attr` and remove any that are no longer present after the plugin evaluates its reactive expression.

## Alternatives Considered

None. The proposed solution seemed ideal to me.

## AI Disclaimer

No AI used for this PR.

### DIFF
--- a/library/src/plugins/attributes/attr.ts
+++ b/library/src/plugins/attributes/attr.ts
@@ -29,6 +29,8 @@ attribute({
       }
     }
 
+    let attributeFilter: string[] = [];
+
     const update = key
       ? () => {
           observer.disconnect()
@@ -41,7 +43,10 @@ attribute({
       : () => {
           observer.disconnect()
           const obj = rx() as Record<string, any>
-          const attributeFilter = Object.keys(obj)
+          for (const key of attributeFilter.filter(key => !obj.hasOwnProperty(key))) {
+            el.removeAttribute(key)
+          }
+          attributeFilter = Object.keys(obj);
           for (const key of attributeFilter) {
             syncAttr(key, obj[key])
           }


bencroker:
Thanks for the PR. It makes sense for the most part, but I’m not 100% sure that this is desirable behaviour. 

What if the attribute in question existed before the `data-attr` value is evaluated, for example? Should it still be removed?

Source: 

```html
<div foo="1" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'bar'">
```

Current behaviour:

```html
<div foo="1" foobar="1" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'bar'">
```

New behaviour:

```html
<div foobar="1" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'bar'">
```

Regaez:
> Thanks for the PR. It makes sense for the most part, but I’m not 100% sure that this is desirable behaviour.
> 
> What if the attribute in question existed before the `data-attr` value is evaluated, for example? Should it still be removed?

Could that be considered "user error"? Datastar has [already established that the order of attributes is important](https://data-star.dev/reference/attributes#attribute-evaluation-order). Per that rule (and with this change), would it not be "expected behaviour" that `foo${$suffix}` would equal `"foo"` when no value for `$suffix` has been previously declared? 

Therefore, if the user wanted to preserve the existing `foo` attribute, they could first set the `$suffix` signal with an appropriate value _before_ declaring `data-attr`, i.e. 

```html
<!-- before data-attr -->
<div foo="1" data-init="$suffix = 'bar'" data-attr="{ [`foo${$suffix}`]: 1 }" >

<!-- after data-attr -->
<div foo="1" foobar="1" data-init="$suffix = 'bar'" data-attr="{ [`foo${$suffix}`]: 1 }" >
```
or, with even a more complex chain of mutations:
```html
<!-- before data-attr -->
<div foo="1" data-signals="{ suffix: 'bar' }" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'qux'">

<!-- after data-attr, before data-init -->
<div foo="1" foobar="1" data-signals="{ suffix: 'bar' }" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'qux'">

<!-- after data-init -->
<div foo="1" fooqux="1" data-signals="{ suffix: 'bar' }" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'qux'">
```
Would that behaviour be consistent with the "attribute evaluation order" rules already laid out in the documentation?

---

That said, we could also consider adding a new modifier to `data-attr` to enable the preferred behaviour, e.g. 

- `__retain`: the element will retain all attribute keys added by `data-attr`, even if they no longer exist in the `data-attr`'s object value. 
_or_
- `__prune`: if any keys are removed from the object given to `data-attr` after it has been evaluated, they will also be removed from the element.

Which of these modifiers to add to the API would depend on the preferred default behaviour.

bencroker:
> Could that be considered "user error"? Datastar has [already established that the order of attributes is important](https://data-star.dev/reference/attributes#attribute-evaluation-order).

The order of attributes determines the order in which they are evaluated, nothing more.

> That said, we could also consider adding a new modifier to `data-attr` to enable the preferred behaviour, e.g.

I’m not keen on adding modifiers unless there’s a clear use-case for it. Is there an actual issue you’re running into in real-world usage, or is this more of a theoretical concern?

Regaez:
> The order of attributes determines the order in which they are evaluated, nothing more.

Yes, but would it not be considered user error that such a  _side effect_ (removing the existing `foo` attribute) is _because the user_ has added `data-attr` with a dependency on a signal that they have not previously declared?

The behaviour (from Datastar's perspective) would be consistent with its stated rules regarding attribute evaluation order. 

> Is there an actual issue you’re running into in real-world usage

Yes, example 1 is a simplified recreation of the scenario I encountered: I wanted to dynamically adjust modifier timing values by way of signals but it resulted in duplicate versions of the attribute being applied to the element, one for every value of the signal.

> I’m not 100% sure that this is desirable behaviour

From a conceptual point of view, does it not make sense that `data-attr` cleans up attributes that it itself has added, based on a reactive expression, when reacting to changes _within that same reactive expression_? 

Is it reasonable to expect that an attribute key that is no longer returned by that expression will therefore no longer be applied to the element? 

Should one view attributes added by `data-attr` as "managed attributes"? Because attributes [can currently be removed by setting the _value_ of the attribute to `null` or `false`](https://github.com/starfederation/datastar/blob/e437da75d2974c3a45ef4de11bed9387d6ab6fdc/library/src/plugins/attributes/attr.ts#L16-L17), implying that `data-attr` does afford some control over the attribute existence and is not solely an "attribute setter". 

But the current API offers no affordance for adjusting Datastar modifier values, since these can _only be declared via the attribute key_.

bencroker:
The stated purpose of `data-attr` is to set the value of any HTML attribute to an expression, and keep it in sync.

You are inferring that removing keys should remove attributes.

> From a conceptual point of view, does it not make sense that `data-attr` cleans up attributes that it itself has added, based on a reactive expression, when reacting to changes within that same reactive expression?

As per my example in https://github.com/starfederation/datastar/pull/1184#issuecomment-4933579503, your PR cleans up `foo` which _already existed_ on the element. Here’s another way of writing it with the same effect, in which `foobar`, an attribute that exists _before_ `data-attr` is evaluated, is removed by your change.

Source: 

```html
<div foobar="1" data-signals:suffix="'bar'" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'baz'">
```

Current result:

```html
<div foobar="1" foobaz="1" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'baz'">
```

New result:

```html
<div foobaz="1" data-attr="{ [`foo${$suffix}`]: 1 }" data-init="$suffix = 'bar'">
```

bencroker:
> Yes, example 1 is a simplified recreation of the scenario I encountered: I wanted to dynamically adjust modifier timing values by way of signals but it resulted in duplicate versions of the attribute being applied to the element, one for every value of the signal.

I think you’re over relying on `data-attr`. Case in point: example 1 can be achieved using a `setTimeout`.

```html
<div data-signals="{ i: 1 }">
  <span data-text="$i"></span>
  <progress max="100" data-bind:i data-init="(tick = () => setTimeout(tick, $i++ * 1000))()"></progress> 
</div>
```

Regaez:
> The stated purpose of data-attr is to set the value of any HTML attribute to an expression, and keep it in sync.

But it doesn't _only_ set attributes, the current version can also _remove them_, [as referenced in my previous comment](https://github.com/starfederation/datastar/pull/1184#issuecomment-4937019864).
 
> You are inferring that removing keys should remove attributes.
>
> As per my example in https://github.com/starfederation/datastar/pull/1184#issuecomment-4933579503, your PR cleans up foo which already existed on the element. 

Yes, that is the intent. Since keys declared within the _value_ of `data-attr` are part of a _reactive expression_ (and the expression can really be _anything_, providing it returns an object[^1]), I believe the attr plugin should also treat keys as potentially mutable, not only the attribute values. 

Granted, this is an increase in scope of what the current attr plugin considers "reactive", but I don't think it's unreasonable. Datastar's API itself includes modifiers that affect attribute behaviour that can _only be configured via attribute keys_, and since Datastar posits itself as a client-side reactive library, it doesn't seem unreasonable to me to be able to adjust those attribute modifiers at runtime on the client. Why should `data-attr` not be the means to do this?

This is why I tried to provide two simple, yet distinct, examples demonstrating how one may apply _different Datastar attributes_ at runtime with `data-attr`, highlighting a gap in the current API.

> Here’s another way of writing it with the same effect, in which foobar, an attribute that exists before data-attr is evaluated, is removed by your change.

Similar behaviour is also reproducible in the _current version_ of Datastar, given the right conditions. For instance, tweaking your previous example slightly:

```html
<div 
  foo="1"
  data-signals="{ isInitialised: false }" 
  data-attr="{ [`foo${$suffix}`]: $isInitialised }" 
  data-init="$suffix = 'bar'; $isInitialised = true;"
>
```
This is 100% valid in Datastar v1.0.2 and this would also result in the existing `foo` attribute being removed. Any existing attribute _could_ be removed, if the key given to `data-attr` happens to match and the value is `false` or `null` at that point in time. 

Is that "user error" or Datastar not behaving as intended? If the former, is that so different to an existing attribute being removed if the key within the `data-attr` happened to, _at one point in time_, also match?

Is it reasonable to reject the behaviour proposed by this PR on the basis that it _might_ remove an existing attribute under specific conditions, when the current Datastar version may do the same thing under its own set of specific conditions?

> I think you’re over relying on data-attr. Case in point: example 1 can be achieved using a setTimeout.

I disagree. This is not equivalent: if the `data-init` attribute, or `<progress>` element, is removed, the `$i` signal will continue to increment indefinitely, i.e. a memory leak. With 

```
data-attr="{ [`data-on-interval__duration.${$i}ms`]: '...' }"
``` 
(or any other attribute with timing modifiers, really) you benefit from automatic cleanup of the interval when it disappears from the DOM, without any convoluted inline IIFE required. There's a clarity of intent and separation of concerns when using a signal to modify the duration, as it leaves the interval's reactive expression untouched and isolated from the orchestration of the timing modifier.

I tried to come up with 2 simple (and visual, for the sake of the live demo) examples so as not to get too bogged down in minutiae, however I fear that it has nonetheless resulted in distraction due to too little code, prompting suggestions for alternative approaches. Were my original Example 1 not simply incrementing `$i` in the interval's expression, but performing some other side effect, and the signal were mutated elsewhere, the "solution" you have suggested would no longer be applicable. However, adjusting the timing modifier with a signal, that remains viable (given the proposed changes in this PR).

[^1]: This, for example, is a valid `data-attr` value: `<div data-attr="let now = Date.now(); return { now };"></div>`

bencroker:
> Granted, this is an increase in scope of what the current attr plugin considers "reactive", but I don't think it's unreasonable. Datastar's API itself includes modifiers that affect attribute behaviour that can _only be configured via attribute keys_, and since Datastar posits itself as a client-side reactive library, it doesn't seem unreasonable to me to be able to adjust those attribute modifiers at runtime on the client. Why should `data-attr` not be the means to do this?

Once again, I think you are inferring that `data-attr` should work how you want it to work rather than fulfil its current scope, and that your example is over relying on the plugin. 

A perfectly reasonable counter-example would be wanting to remove a key but not the attribute. Or wanting to remove the entire `data-attr` attribute but not the attributes that it was keeping in sync.

A `__prune` modifier might be worth considering, but it adds extra logic, docs, and more cleanup code than what you currently have.

> This is 100% valid in Datastar v1.0.2 and this would also result in the existing `foo` attribute being removed. Any existing attribute _could_ be removed, if the key given to `data-attr` happens to match and the value is `false` or `null` at that point in time. 
>
> Is that "user error" or Datastar not behaving as intended? If the former, is that so different to an existing attribute being removed if the key within the `data-attr` happened to, _at one point in time_, also match?

This is intended behaviour. Setting the value to `false`/`null` removes the attribute, which is important to be able to do for boolean attributes.

Regaez:
> This is intended behaviour. Setting the value to false/null removes the attribute, which is important to be able to do for boolean attributes.

So it is "user error" then, that the existing `foo` attribute is removed once `data-attr` is evaluated.

> A perfectly reasonable counter-example would be wanting to remove a key but not the attribute. Or wanting to remove the entire `data-attr` attribute but not the attributes that it was keeping in sync.

Hence my suggestion to add a modifier to enable the user to opt-in to the type of behaviour they would prefer. I have no issue with the current behaviour remaining the default.

> A `__prune` modifier might be worth considering, but it adds extra logic, docs, and more cleanup code than what you currently have.

I am more than happy to discuss the details of such a modifier and contribute these additional changes (including relevant documentation), as I believe this functionality would be a valuable addition to the attr plugin.

bencroker:
> So it is "user error" then, that the existing `foo` attribute is removed once `data-attr` is evaluated.

If that is not what the user intended to do then yes, it is user error.

I’ll take a look at what it will entail and will discuss with the team.

Regaez:
> I’ll take a look at what it will entail and will discuss with the team.

Thank you.

bencroker:
We discussed internally and decided that removing attributes should _not_ be part of `data-attr`’s cleanup. Its purpose is to sync attribute values, and remove them when the value is `false/null`. You’re trying to add scope to the plugin, rather than considering alternative solutions.

